PR #65, opened 2011-09-23
added interfaces for ManagerRegistry, as well as a default implementation

lsmith77:
the creation of the ConnectionRegistry was mostly derived from trying to maintain as much BC as possible with the RegistryInterface currently shipped with Symfony2 but running into https://bugs.php.net/bug.php?id=43200

all implementation (ORM/ODM) will need to match the changes to the ClassMetadataFactory interface

ORM, PHPCR, CouchDB have been upgraded already

### DIFF
--- a/lib/Doctrine/Common/Persistence/AbstractManagerRegistry.php
+++ b/lib/Doctrine/Common/Persistence/AbstractManagerRegistry.php
@@ -54,6 +54,8 @@ abstract class AbstractManagerRegistry implements ManagerRegistry, ConnectionReg
     /**
      * Fetches/creates the given services
      *
+     * A service in this context is connection or a manager instance
+     *
      * @param string $name name of the service
      * @return object instance of the given service
      */
@@ -62,6 +64,8 @@ abstract class AbstractManagerRegistry implements ManagerRegistry, ConnectionReg
     /**
      * Resets the given services
      *
+     * A service in this context is connection or a manager instance
+     * 
      * @param string $name name of the service
      * @return void
      */


lib/Doctrine/Common/Persistence/AbstractManagerRegistry.php
```diff
@@ -0,0 +1,218 @@
...
+    /**
+     * @inheritdoc
+     */
+    public function getRepository($persistentObjectName, $persistentManagerName = null)
+    {
+        return $this->getManager($persistentManagerName)->getRepository($persistentObjectName);
+    }
+
+    /**
+     * @inheritdoc
```
ogizanagi: The `AbstractManagerRegistry:resetManager` implementation does not respect the `Doctrine\Common\Persistence\ManagerRegistry` interface and does not return the new instance of `\Doctrine\Common\Persistence\ObjectManager`. Is it expected ?

stof: @ogizanagi if you find an issue, it is much better to open a ticket rather than commenting on a PR merged nearly 5 years ago. Maintainers are not monitoring old PR to check for new discussions (I saw your comment because of the notification email, but this is luck, as I'm not looking at all notifications).


Regarding the issue, I suggest changing the phpdoc of the interface, because the code never returned an instance, and so all code written in 5 years is built around the fact that resetManager will reset the manager retrieved on the next call to getManager

ogizanagi: Thanks @stof for your answer. I thought a single comment to get an explanation on the original PR was enough in the first place, but I totally understand and I know what to do next time :)

PR #153, opened 2012-06-15
[3.0] New Event system

guilhermeblanco:


Ocramius:
Closing: after 3 years, I think we might first look at how the ecosystem may have evolved before going and building our own (again)

PR #317, opened 2014-02-21
Implemented an ObjectPersisterInterface for entity/object storage

linaori:
### Why is this useful?

Instead of using the `repositoryClass`, we use 'Repository as a Service'. This means that we inject our repository service instead of doing `$em->getRepository('MyEntity')`, this provides typehinting and autocomplete in IDEs. This introduces a minor inconvenience, we need to inject an `EntityManagerInterface` to provide persist/flush. We don't want the entire `EntityManagerInterface` capabilities just to store an object.

Using the `ObjectPersisterInterface` we can "hide" the `EntityManagerInterface` and only let the code know we have the persist and flush methods available. By default this is implemented on the `EntityManager`, but is also possible to be mocked or replaced by a custom implementation (rotating entity managers?).
#### Example

``` php

class MyService
{
    private $op;
    private $es;

    public function __construct(ObjectPersisterInterface $op, MyEntityService $es)
    {
        $this->op = $op;
        $this->es = $es;
    }

    public function doSomething($value)
    {
        $entity = $this->es->findByValue($value)->setValue('something');
        $this->op->flush();
    }

    public function doSomethingElse()
    {
        $entity = new MyEntity();
        $this->op->persist($entity);
        $this->op->flush();
    }
}

```

### DIFF
--- a/lib/Doctrine/Common/Persistence/ObjectManager.php
+++ b/lib/Doctrine/Common/Persistence/ObjectManager.php
@@ -27,7 +27,7 @@ namespace Doctrine\Common\Persistence;
  * @author Benjamin Eberlei <kontakt@beberlei.de>
  * @author Jonathan Wage <jonwage@gmail.com>
  */
-interface ObjectManager extends ObjectManagerPersister
+interface ObjectManager extends ObjectPersisterInterface
 {
     /**
      * Finds an object by its identifier.


WouterJ:
Given that a core team member thinks this kinda makes sense, can this please be reconsidered? Having to create an interface doing this in all packages seems strange...

Ocramius:
@WouterJ looked at it again, and I don't think we should re-evaluate this before having a plan for new APIs. For now, composition is enough.

PR #319, opened 2014-03-22
Add last modified time for metadata

c960657:
This PR is part 1 of 2. Part 1 is for the doctrine/doctrine2 repository.

Read more in https://github.com/doctrine/doctrine2/pull/986


Ocramius:
After much much thinking about this (got back to this issue a dozen times while reviewing other issues), I do not think that the current architecture (with half the metadata API here and half in the ORM or other implementations) is well suited for this sort of feature.

We basically risk building something very complex and possibly hard to debug, and only for the benefit of dev environments.

I suggest simply using a file watcher and clearing the metadata manually, for now.

Sorry for all the time I've wasted with technical review and comments, @c960657, I simply don't think the feature is worth the additional complexity for now :-(

A different approach could be to turn the common metadata into an "envelope" object in which child drivers can write/read other instances, but that's something @guilhermeblanco is working on right now.

I'll re-assign the issue to him in order for him to have a look at it and see if there is something worth implementing/porting to his current rework.

c960657:
Thanks for the feedback. I appreciate getting a final decision with a clear stance (allthough a little late :-), rather than having the PR sitting around forever, because nobody wants to disappoint anyone by closing it.

In case anybody is interested, here is a what I did instead:

```
$lastModifyUts = 0;
foreach ($app['orm.em.options']['mappings'] as $mapping) {
    foreach (scandir($mapping['path']) as $basename) {
        $file = $mapping['path'] . '/' . $basename;
        if (is_file($file)) {
            $lastModifyUts = max($lastModifyUts, filemtime($file));
        }
    }
}
$doctrineOrmOptions['orm.em.options']['cache_namespace'] = $lastModifyUts;
```

Ocramius:
Closing as per discussion in doctrine/doctrine2#986

PR #325, opened 2014-07-04
Create proxy cache dir with 777

hboomsma:
All other directories in Doctrine use 777 but this one. This causes access problems when clearing the cache when the first request was done via a webbrowser and the files are owned by apache/www-data. Now a console / deployment user can not clear the cache any longer. The problem also works the other way around.


stof:
@Ocramius no it is not. ``mkdir`` applies the umask on the permissions you pass to it. And on all distributions, the default umask is ``0022``, making it a safe permission.
The difference is that by using 0777 in the code, you give control to the user (in case they cannot have their webserver and the CLI using the same user and they cannot use ACLs, thus having to rely on group permissions).

PR #333, opened 2014-08-11
update the object state when calling PreUpdateEventArgs::setNewValue()

lsmith77:
related to http://www.doctrine-project.org/jira/browse/DDC-3248 (guess ORM would then also need to be updated). or maybe it should only be done in specialized classes.


Ocramius:
Closing as incomplete.

PR #335, opened 2014-08-28
add base events class

jdeniau:
Hi,

All doctrine projects implements events. A lot of them are similars, but there is no base class defining common Events.

I tried to gathen them all in this PR. I looked into `CouchDb-orm`, `doctrine-orm`, `doctrine-mongodb-odm` and `phpcr-odm`. 

I only added events present in all packages, but CouchDB is the only one to not implement some events implemented in every other packages (`postPersist`, `loadClassmetadata`, `preFlush`, etc.).


jdeniau:
So, what is the state of this ? 

It is just a standardization of event names used in every doctrine packages.

It really helps external packages development as the event names are constants in the common package. If the event names are stored in the consumer, you need to : add every consumer as a dependency (just impossible), or test directly the strings (really not safe).

Majkl578:
Hello,
we've recently split Event Manager component into a separate [doctrine/event-manager](https://github.com/doctrine/event-manager) repository.
Thanks.

PR #337, opened 2014-09-03
[WIP] added a test for identifier getter in a trait

dorongutman:
When the identifier getter method is in a trait, the proxy generation ignores it and treats it as a lazy loading field, causing the $proxyInstance->getId() to initialize the object.

This is a test that proves that.


Ocramius:
Closing. A solution to this (without going completely crazy) would be doctrine/doctrine2#1241, but we're far from implementing it.

PR #338, opened 2014-09-22
Short gettter should work with associations

goetas:


goetas:
At the moment I do not have time to work on it... if #1241 will be merged, the problem will be solved in better way

PR #342, opened 2014-11-11
Class metadata loading fallback hook

Ocramius:
Enables loading class metadata when loading failed (in subclasses).

See doctrine/doctrine2#385

Note that I preferred inheritance to avoid having an `EventManager` in here, though it may be a perfectly acceptable dependency in this case, therefore it is up for discussion.


Ocramius:
Merging as per @guilhermeblanco's sign-off.

PR #343, opened 2014-11-12
new interface for mapped property names interface

fabiocarneiro:
The current implementation of ClassMetadata does not provide any method to return the mapped property names, and since we can't add it to the current ClassMetadata interface and the DoctrineObject hydrator does not use the implementation, its necessary to have another interface to make sure the new feature of ClassMetadata implemetation is available and keep both the DoctrineObject and ClassMetadata implementation backwards compatible.


Ocramius:
Closing: the metadata API is being reworked by @guilhermeblanco, and introducing new interfaces now will just lead to bigger BC breaks later on

PR #345, opened 2014-12-25
Cleanup: class loader deprecation and simplification

Ocramius:
Time to get rid of this thing: marking it as deprecated.


FoxxMD:
So what should I be using instead of `ClassLoader`?

PR #347, opened 2014-12-31
Class loader: skip non-existing files and files not containing the requested class

Ocramius:
See https://github.com/doctrine/common/pull/345#discussion_r22356775

### DIFF
--- a/lib/Doctrine/Common/ClassLoader.php
+++ b/lib/Doctrine/Common/ClassLoader.php
@@ -175,10 +175,6 @@ class ClassLoader
      */
     public function loadClass($className)
     {
-        if ($this->namespace !== null && strpos($className, $this->namespace.$this->namespaceSeparator) !== 0) {
-            return false;
-        }
-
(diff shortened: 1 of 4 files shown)


Ocramius:
@stof factored in all the feedback. trait support now built in and it shouldn't be possible to use this thing to cause fatal errors anymore.

lib/Doctrine/Common/ClassLoader.php
```diff
@@ -233,21 +237,7 @@ public function canLoadClass($className)
      */
     public static function classExists($className)
     {
-        if (class_exists($className, false) || interface_exists($className, false)) {
-            return true;
-        }
-
-        foreach (spl_autoload_functions() as $loader) {
```
deeky666: Isn't that necessary anymore?

Ocramius: No, it was just emulating `class_exists()` internally

PR #349, opened 2015-01-20
Add autoload to composer.json

pgddevil:


Ocramius:
@pgddevil we don't support `2.0.x` anymore: why are you suggesting this change here?

pgddevil:
because I have not yet upgraded and wanted to use composer.  quite useless to have added a composer.json without tagging and without the autoload.

pgddevil:
However I can declare the package in my own composer.json if you don't want to go further with the autoload add-on.

fprochazka:
@pgddevil you can add the autoloader definition to autoloader section of your project and it will also work.

pgddevil:
Well, I know I can do it another way.  I wanted to do it the clean way.  I've already fixed my situation.  It's up to you to decide to merge or not.

Ocramius:
Yeah, you should handle autoloading yourself here.

2.0.0 is pre-composer (composer didn't even exist), and it's not a good idea to exhume a dead release.

Consider upgrading to a newer version of the library.

PR #350, opened 2015-01-23
Made PATTERN_MATCH_ID_METHOD more flexible

4alexandr:
PATTERN_MATCH_ID_METHOD match any valid function declaration, example:
function getId(){return$this->id;}

### DIFF
--- a/lib/Doctrine/Common/Proxy/ProxyGenerator.php
+++ b/lib/Doctrine/Common/Proxy/ProxyGenerator.php
@@ -37,7 +37,7 @@ class ProxyGenerator
      * Used to match very simple id methods that don't need
      * to be decorated since the identifier is known.
      */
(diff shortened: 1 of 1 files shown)


Ocramius:
@4alexandr requires test case.

4alexandr:
@Ocramius Sorry, i don't know, how to write tests for this change. I can test this only manually.
Old pattern can match 
```php
    public function getId()
    {
        return $this->id;
    }
```
but can't
```php
    public function getId(){
        return $this->id;
    }
```
In commit i change non-required whitespaces detection in function declaration.

Ocramius:
@4alexandr look at existing tests for the proxy generator ;-)

Ocramius:
Is also fixed by something like doctrine/doctrine2#1241 (linking for reference - this partial fix is fine for now)

Ocramius:
Related (not blocking) doctrine/doctrine2#1241

Ocramius:
Checked again, and writing a test case for this is a huge annoyance for little benefit. Merging as-is.

PR #351, opened 2015-02-01
[WIP] Improved conversion for aliased classnames

phansys:
Improved conversion for aliased classnames

**Before:**
```Doctrine\Tests\Common\Persistence\Mapping:ChildEntity:Foo``` resolves (silently) to ```Doctrine\Tests\Common\Persistence\Mapping\ChildEntity```

**After:**
```Doctrine\Tests\Common\Persistence\Mapping:ChildEntity:Foo``` resolves to ```Doctrine\Tests\Common\Persistence\Mapping\ChildEntity:Foo``` in order to get validation in lower layers.

| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | none
| License       | MIT
| Doc PR        | none

TODO:
- [x] Add tests for AbstractClassMetadataFactory::getMetadataFor()
- [x] Add tests for MappingDriver::isTransient()
- [x] Add tests for AbstractManagerRegistry::getManagerForClass()

### DIFF
--- a/lib/Doctrine/Common/Persistence/AbstractManagerRegistry.php
+++ b/lib/Doctrine/Common/Persistence/AbstractManagerRegistry.php
@@ -190,7 +190,10 @@ abstract class AbstractManagerRegistry implements ManagerRegistry
     {
         // Check for namespace alias
         if (strpos($class, ':') !== false) {
-            list($namespaceAlias, $simpleClassName) = explode(':', $class);
+            if (2 < count($classNameAlias = explode(':', $class))) {
+                throw new \InvalidArgumentException(sprintf('Invalid classname "%s".', $class));
+            }
+            list($namespaceAlias, $simpleClassName) = $classNameAlias;
             $class = $this->getAliasNamespace($namespaceAlias) . '\\' . $simpleClassName;
         }
 
--- a/lib/Doctrine/Common/Persistence/AbstractManagerRegistry.php
+++ b/lib/Doctrine/Common/Persistence/AbstractManagerRegistry.php
@@ -190,10 +190,7 @@ abstract class AbstractManagerRegistry implements ManagerRegistry
     {
         // Check for namespace alias
         if (strpos($class, ':') !== false) {
-            if (2 < count($classNameAlias = explode(':', $class))) {
-                throw new \InvalidArgumentException(sprintf('Invalid classname "%s".', $class));
-            }
-            list($namespaceAlias, $simpleClassName) = $classNameAlias;
+            list($namespaceAlias, $simpleClassName) = explode(':', $class, 2);
             $class = $this->getAliasNamespace($namespaceAlias) . '\\' . $simpleClassName;
         }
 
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

**Comment by Ocramius** [comment]
> `MyEntityNamespace:Foo:Bar` resolves (silently) to `MyEntityNamespace:Foo`

`MyEntityNamespace:Foo` is an invalid class name (will fail to autoload) anyway, no?

File: lib/Doctrine/Common/Persistence/AbstractManagerRegistry.php [hunk 1]
```diff
@@ -190,7 +190,10 @@ public function getManagerForClass($class)
     {
         // Check for namespace alias
         if (strpos($class, ':') !== false) {
-            list($namespaceAlias, $simpleClassName) = explode(':', $class);
+            if (2 < count($classNameAlias = explode(':', $class))) {
```
Ocramius: Note that `explode()` has a third `$limit` argument (see http://3v4l.org/as2I7), maybe that's a better solution here...

phansys: Yes @Ocramius, thank you for the tip.
I updated the PR with this change.

phansys:
Yes @Ocramius, of course ```MyEntityNamespace:Foo``` is an invalid CN. Sorry for the misunderstood, I did mean that these parts of the given alias were catched while the convertion to the real CN (```MyEntityNamespace\Foo``` in my example). BTW, I updated the examples in the PR message in order to reflect the current status.

Now I need some hint in order to finish the 2nd and 3rd item.
Thank you.

Ocramius:
Tests for `isTransient` and for the `AbstractManagerRegistry` should also be provided

phansys:
Sure, this is the reason of the updated title with "[WIP]" prepended; but I tried to add a test for ```isTransient()``` without success because the driver isn't validating the classname.

Test:
```php
public function testInvalidClassIsTransient()
{
    $this->setExpectedException(
        'Doctrine\Common\Persistence\Mapping\MappingException',
        'Class \'Doctrine\Tests\Common\Persistence\Mapping\ChildEntity:Foo\' does not exist'
    );

    $this->cmf->isTransient('prefix:ChildEntity:Foo');
}
```

results in:
```
There was 1 failure:

1) Doctrine\Tests\Common\Persistence\Mapping\ClassMetadataFactoryTest::testInvalidClassIsTransient
Failed asserting that exception of type "Doctrine\Common\Persistence\Mapping\MappingException" is thrown.
```

Ocramius:
@phansys should just report `true`, tbh

phansys:
Perfect @Ocramius, I will add a test for this case with that assert.
What about ```AbstractManagerRegistry```? Do you have in mind some existent test in which this one could be appended?

Ocramius:
@phansys thanks for providing tests. I'll review and merge manually after fixing nitpickery myself.

phansys:
You're welcome @Ocramius!

Ocramius:
Merged, thanks!

PR #352, opened 2015-02-02
SymfonyFileLocator: DIRECTORY_SEPARATOR instead of slash

bartekgruszka:
It could cause some odd behaviour especially on Windows.


Ocramius:
> could cause

This means that we need a test case for a windows system ;-)

Ocramius:
Closing due to missing response.

PR #353, opened 2015-02-09
Support for Variadic arguments (PHP 5.6 - Argument unpacking)

ruudk:
See http://www.doctrine-project.org/jira/browse/DCOM-272

It converts
```php
    /**
     * @param ...$types
     */
    public function addType(...$types)
    {
    }
```

into
```php
    /**
     * {@inheritDoc}
     */
    public function addType(...$types)
    {

        $this->__initializer__ && $this->__initializer__->__invoke($this, 'addType', array($types));

        return parent::addType(...$types);
    }
```

This works but I don't know if the `__invoke` call is correct.

### DIFF
--- a/tests/Doctrine/Tests/Common/Proxy/ProxyClassGeneratorTest.php
+++ b/tests/Doctrine/Tests/Common/Proxy/ProxyClassGeneratorTest.php
@@ -163,6 +163,27 @@ class ProxyClassGeneratorTest extends PHPUnit_Framework_TestCase
         $this->assertEquals(1, substr_count($classCode, 'call(callable $foo)'));
     }
 
+    public function testClassWithVariadicArgumentOnProxiedMethod()
+    {
+        if (PHP_VERSION_ID < 50600) {
+            $this->markTestSkipped('`...` is only supported in PHP >=5.6.0');
+        }
+
+        if (!class_exists('Doctrine\Tests\Common\ProxyProxy\__CG__\VariadicTypeHintClass', false)) {
+            $className = 'Doctrine\Tests\Common\Proxy\VariadicTypeHintClass';
+            $metadata = $this->createClassMetadata($className, array('id'));
+
+            $proxyGenerator = new ProxyGenerator(__DIR__ . '/generated', __NAMESPACE__ . 'Proxy', true);
+            $this->generateAndRequire($proxyGenerator, $metadata);
+        }
+
+        $classCode = file_get_contents(__DIR__ . '/generated/__CG__DoctrineTestsCommonProxyVariadicTypeHintClass.php');
+
+        $this->assertEquals(1, substr_count($classCode, 'function addType(...$types)'));
+        $this->assertEquals(1, substr_count($classCode, '__invoke($this, \'addType\', array($types))'));
+        $this->assertEquals(1, substr_count($classCode, 'parent::addType(...$types)'));
+    }
+
     public function testClassWithInvalidTypeHintOnProxiedMethod()
     {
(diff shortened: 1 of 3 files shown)


tests/Doctrine/Tests/Common/Proxy/VariadicTypeHintClass.php
```diff
@@ -0,0 +1,16 @@
...
+/**
+ * Test asset class
+ */
+class VariadicTypeHintClass
+{
+    /**
+     * @param ...$types
+     */
+    public function addType(...$types)
+    {
```
Ocramius: You really want to test the return value as well

tests/Doctrine/Tests/Common/Proxy/ProxyClassGeneratorTest.php
```diff
@@ -163,6 +163,27 @@ public function testClassWithCallableTypeHintOnProxiedMethod()
...
+
+            $proxyGenerator = new ProxyGenerator(__DIR__ . '/generated', __NAMESPACE__ . 'Proxy', true);
+            $this->generateAndRequire($proxyGenerator, $metadata);
+        }
+
+        $classCode = file_get_contents(__DIR__ . '/generated/__CG__DoctrineTestsCommonProxyVariadicTypeHintClass.php');
+
+        $this->assertEquals(1, substr_count($classCode, 'function addType(...$types)'));
+        $this->assertEquals(1, substr_count($classCode, '__invoke($this, \'addType\', array($types))'));
+        $this->assertEquals(1, substr_count($classCode, 'parent::addType(...$types)'));
```
Ocramius: Probably requires an integration test as well

stof:
looks good to me. the invoker calls looks good to me. It makes sense to consider the internal behavior of arguments (i.e. a ``$types`` array argument) rather than the external one (a variadic list of types)

tests/Doctrine/Tests/Common/Proxy/VariadicTypeHintClass.php
ruudk: How do you mean?

tests/Doctrine/Tests/Common/Proxy/ProxyClassGeneratorTest.php
ruudk: Sure, do you have an example?

Ocramius: I think https://github.com/doctrine/common/blob/56360cc13bfd13893731c6c4d1d82005a61d7694/tests/Doctrine/Tests/Common/Proxy/ProxyLogicTest.php does it

tests/Doctrine/Tests/Common/Proxy/VariadicTypeHintClass.php
Ocramius: Relevant for the proxy logic test (just to test that a method was called with the right amount of args)

Ocramius:
Just misses the functional test, then I'm merging :)

ruudk:
I'll work on it later :)

ruudk:
@Ocramius Hmm. I'm kinda stuck. Don't know how to do this exactly...

Ocramius:
I've manually merged this at doctrine/common@51af551eaab389a0f53b13afdbc9c7f0e431de88 after adding the functional tests (see commit for reference)

ruudk:
@Ocramius Thanks, you are the best :D

PR #355, opened 2015-02-20
Added Magic Function For EntityManger

anilGupta:
- findArrayResultBy
  - findOneArrayResultBy

   [pullrequest] (https://github.com/doctrine/doctrine2/pull/1306)


stof:
changing the interface is a huge BC break, so this cannot be accepted (and it is not even consistent with your PR to the ORM btw).

And btw, the functions you are adding here are not magic ones.

PR #356, opened 2015-02-23
Identifier getters now found in traits when generating proxies.

tPl0ch:
Related to http://www.doctrine-project.org/jira/browse/DDC-2556

TODO: Add test cases

### DIFF
--- a/lib/Doctrine/Common/Proxy/ProxyGenerator.php
+++ b/lib/Doctrine/Common/Proxy/ProxyGenerator.php
@@ -847,13 +847,16 @@ EOT;
         );
 
         if ($cheapCheck) {
-            $code = file($method->getDeclaringClass()->getFileName());
+            $methodFile = $method->getDeclaringClass()->getFileName();
+            $code = file($methodFile);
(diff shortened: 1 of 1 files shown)


Ocramius:
Note: I'd rather NOT have trait support for this rather than adding this bunch of very fragile complexity.

Additionally, see doctrine/doctrine2#1241, which would solve this upfront.

tPl0ch:
@Ocramius So waiting for 3.0 is the only chance of getting our traits back, am I right?

Ocramius:
Probably, yeah... I could also accept this hotfix if accurately tested (all npaths)

enumag:
The problem is that `ReflectionMethod::getDeclaringClass()` does not return the trait where the method was declared. In my opinion the correct solution would be to use a [helper method](https://github.com/nette/reflection/blob/master/src/Reflection/Helpers.php#L24-L32) which fixes that. This might also be relevant to some other places where `ReflectionMethod::getDeclaringClass()` is used.

Ocramius:
Closing: the complexity of this fix, as well as the depth of the bug (touching all the weird parts of traits) are not worth fixing, IMO.

Ocramius:
Note: not a `won't fix`, but a `can't fix`.

PR #357, opened 2015-02-24
Custom namespace separators for SymfonyFileLocator

scaytrase:
As a Symofony user I'm tired of storaging dozens of yml metadata files within `config\doctrine` folder as a plain list. Current implementation of `SymfonyFileLocator` fails to loopback class name due to automatic convertation extra classname separators behind default `Entity` prefix to dots, so placing yaml metadata files into nested folders does not work at all.

This also covers some other people needs as I see by 
http://stackoverflow.com/questions/14530782/symfony2-doctrine-entity-subdirectories
http://stackoverflow.com/questions/20998622/is-anybody-knows-how-organize-orm-yml-for-symfony-2-x

I've implemented custom namespace separator value to be substitued when `SymfonyFileLocator` lookup classname and converts filename to classname, allowing to store entity mapping with the same structure are placed, i.e with `DIRECTORY_SEPARATOR` used

    Foo\Bar => config/doctrine/Foo/Bar.orm.yml
    Baz => config/doctrine/Baz.orm.yml

etc.

I've added a bunch of tests to be sure I've not broken something, with two examples of usage. 
 * Classic - dot (`.`) separator. This is default for keeping BC
 * `DIRECTORY_SEPARATOR` - for representing namespaces as `PSR-4` / `PSR-0` structure with according prefix

Local (Windows, PHP 5.6) and Travis tests are passed 

https://travis-ci.org/doctrine/common/builds/52018360


lib/Doctrine/Common/Persistence/Mapping/Driver/SymfonyFileLocator.php
```diff
@@ -54,15 +54,24 @@ class SymfonyFileLocator implements FileLocator
...
      * @param array       $prefixes
      * @param string|null $fileExtension
+     * @param string      $nsSeparator
      */
-    public function __construct(array $prefixes, $fileExtension = null)
+    public function __construct(array $prefixes, $fileExtension = null, $nsSeparator = '.')
     {
         $this->addNamespacePrefixes($prefixes);
         $this->fileExtension = $fileExtension;
+        $this->nsSeparator = $nsSeparator;
```
Ocramius: Cast to `(string)`. I'd also disallow an empty separator

Ocramius:
This looks good to me except for the constructor argument :-)

lib/Doctrine/Common/Persistence/Mapping/Driver/SymfonyFileLocator.php
```diff
@@ -54,15 +54,24 @@ class SymfonyFileLocator implements FileLocator
...
+     * @var string
+     */
+    protected $nsSeparator;
+
+    /**
      * Constructor.
      *
      * @param array       $prefixes
      * @param string|null $fileExtension
+     * @param string      $nsSeparator
```
Ocramius: Please document this in the docblock

scaytrase: Is class property documentation not enough? Or should I move it here or duplicate?

```php
    /**
     * Represents PHP namespace delimiters when looking for files
     *
     * @var string
     */
    protected $nsSeparator;
```

Ocramius: What's visible to the consumer is the constructor signature

stof:
Using an underscore is a bad idea, because it is valid inside namespaces and class names, and so you will have several class names mapping to the same file. Only characters not allowed in PHP identifiers should be allowed for the namespace separator.

lib/Doctrine/Common/Persistence/Mapping/Driver/SymfonyFileLocator.php
scaytrase: Something like 

```php
$this->nsSeparator = (string) $nsSeparator ?: '.'
```

?

lib/Doctrine/Common/Persistence/Mapping/Driver/SymfonyFileLocator.php
```diff
@@ -54,15 +54,24 @@ class SymfonyFileLocator implements FileLocator
     protected $fileExtension;
 
     /**
+     * Represents PHP namespace delimiters when looking for files
+     *
+     * @var string
+     */
+    protected $nsSeparator;
```
stof: please make it private.
Anything protected is an extension point for which we must maintain BC, so it is better to use private visibility when introducing new properties/methods.
Increasing the visibility is easy (but may not be desirable which is why we are always asking for justification in this case to evaluate whether it is the right way), but the opposite (turning a protected member into a private one) is a BC break (and so is forbidden except in major versions)

lib/Doctrine/Common/Persistence/Mapping/Driver/SymfonyFileLocator.php
Ocramius: No, invalid input should cause a hard failure

lib/Doctrine/Common/Persistence/Mapping/Driver/SymfonyFileLocator.php
Ocramius: Agree

lib/Doctrine/Common/Persistence/Mapping/Driver/SymfonyFileLocator.php
```diff
@@ -172,9 +181,21 @@ public function getAllClassNames($globalBasename = null)
 
                     // NOTE: All files found here means classes are not transient!
                     if (isset($this->prefixes[$path])) {
-                        $classes[] = $this->prefixes[$path].'\\'.str_replace('.', '\\', $fileName);
+
+                        // Calculate subspace as a relative path from basepath to file path
+                        $subspace = substr(realpath($file->getPath()), strlen(realpath($path)));
```
stof: what does ``subspace`` mean here ? The name is not clear

tests/Doctrine/Tests/Common/Persistence/Mapping/SymfonyFileLocatorTest.php
```diff
@@ -62,6 +62,90 @@ public function testGetAllClassNames()
         $this->assertEquals(array("Foo\\stdClass"), $locator->getAllClassNames("global"));
     }
 
+
+    public function customNamespaceSeparatorProvider()
+    {
+        return array(
+            'directory separator' => array(DIRECTORY_SEPARATOR, "/_custom_ns/dir"),
+            'underscore separator' => array('_', "/_custom_ns/underscore"),
```
stof: this case is a broken setup, so the testsuite should not advocate it (see my comment for the explanation)

Ocramius: TBH, I don't see a problem with supporting `_` here. Even with classes using `_` as pseudo-namespace-separator, the chances of collisions are very low and impossible with default values

lib/Doctrine/Common/Persistence/Mapping/Driver/SymfonyFileLocator.php
scaytrase: I refer it as class nested namspace relative to given namespace prefix. I really don't know how to refer a part of namespace chain properly in english

tests/Doctrine/Tests/Common/Persistence/Mapping/SymfonyFileLocatorTest.php
scaytrase: I've re-read PSR-0 again and realised that underscore notation should affect only `Class_Name` notation and should not affect the `name_space` part, so it's indeed a bad setup example, but not sure if the should forcibly prohibit underscores for use

scaytrase:
I've updated the PR accroding to your notices, except one about `subspace` as I have no idea how to name it properly. Removed undescore references from PR

tests/Doctrine/Tests/Common/Persistence/Mapping/SymfonyFileLocatorTest.php
stof: And PSR-4 removed the special meaning for underscores in the class name too

scaytrase: Yes, I know this. So now reference to underscore is removed from everywhere. It's only possibly to fault when end user manually sets underscore as namespace separator when storing mappings.

scaytrase:
I've rewrited `subspace` to `nsSuffix` and removed unused checks (they really do nothing, because if `nsSeparator` is not used within suffix, and suffix is not empty - then backward lookup (`findMappingFile()`) will fail anyway, so I'ts also broken setup checks)

Ocramius:
@scaytrase merged, thanks!

scaytrase:
Thank you, @Ocramius and @stof. I'll try if I can make this configurable on higher levels

mvrhov:
This PR introduced a regression for my usecase. I have a different mapping in a subdirectory for a "backup" database. Now the different mappings are automatically merged with original mappings and the subdirectory is added as a namespace. 
Also it seems that the subdirectory uses a / character in a namespace.

scaytrase:
@mvrhov Previous implementations should convert `./backup/someClass.yml` to `someClass` classname, which fail unless you have `./someClass.yml` at the root mapping directory (because `findMappingInfo` fails to find mapping file for class). In case you have both files it would silently work and use `./someClass.yml` file for mapping.

Current implementation just converts `./backup/someClass.yml` to `backup\someClass` classname, (on Windows, i think, because here `DIRECTORY_SEPARATOR` is equal to namespace separator) which instantly fails, because not backup.someClass.yml present for it.

Initially I have the check for subpath, which do not use any `nsSeparator` within, but later I've remove it as a check for broken setup.

This is also reason why you have odd classes in the mapping - backup and current mappings are mapped to the different classes now.

mvrhov:
@scaytrase `./backup/someClass.yml` is converted into `NAMESPACE/backup\someClass.yml` on linux. Not counting the BC break.

KingCrunch:
@scaytrase This is a BC-break and it already has a victim :wink: 

```
                                                           
  [Doctrine\Common\Persistence\Mapping\MappingException]   
  Class 'FOS\UserBundle\Entity/model\User' does not exist  
                                                           


Exception trace:
 () at /vagrant/vendor/doctrine/common/lib/Doctrine/Common/Persistence/Mapping/MappingException.php:96
 Doctrine\Common\Persistence\Mapping\MappingException::nonExistingClass() at /vagrant/vendor/doctrine/common/lib/Doctrine/Common/Persistence/Mapping/RuntimeReflectionService.php:41
 Doctrine\Common\Persistence\Mapping\RuntimeReflectionService->getParentClasses() at /vagrant/vendor/doctrine/common/lib/Doctrine/Common/Persistence/Mapping/AbstractClassMetadataFactory.php:281
 Doctrine\Common\Persistence\Mapping\AbstractClassMetadataFactory->getParentClasses() at /vagrant/vendor/doctrine/common/lib/Doctrine/Common/Persistence/Mapping/AbstractClassMetadataFactory.php:311
 Doctrine\Common\Persistence\Mapping\AbstractClassMetadataFactory->loadMetadata() at /vagrant/vendor/doctrine/orm/lib/Doctrine/ORM/Mapping/ClassMetadataFactory.php:78
 Doctrine\ORM\Mapping\ClassMetadataFactory->loadMetadata() at /vagrant/vendor/doctrine/common/lib/Doctrine/Common/Persistence/Mapping/AbstractClassMetadataFactory.php:216
 Doctrine\Common\Persistence\Mapping\AbstractClassMetadataFactory->getMetadataFor() at /vagrant/vendor/doctrine/common/lib/Doctrine/Common/Persistence/Mapping/AbstractClassMetadataFactory.php:115
 Doctrine\Common\Persistence\Mapping\AbstractClassMetadataFactory->getAllMetadata() at /vagrant/vendor/doctrine/data-fixtures/lib/Doctrine/Common/DataFixtures/Purger/ORMPurger.php:102
 Doctrine\Common\DataFixtures\Purger\ORMPurger->purge() at /vagrant/vendor/doctrine/data-fixtures/lib/Doctrine/Common/DataFixtures/Executor/AbstractExecutor.php:136
 Doctrine\Common\DataFixtures\Executor\AbstractExecutor->purge() at /vagrant/vendor/doctrine/data-fixtures/lib/Doctrine/Common/DataFixtures/Executor/ORMExecutor.php:80
```

Simply said: This breaks `FOSUserBUndle`

Ocramius:
As per @guilhermeblanco's findings, seems like FOSUB was using it wrong in
first place :-\

Marco Pivetta

http://twitter.com/Ocramius

http://ocramius.github.com/

KingCrunch:
@Ocramius Tbh I just read through all comments. Debugging this issue was hard enough -_- I can confirm, that in the latest master it doesn't appear anymore. Although it doesn't affect me anymore it looks like a BC-break to me :X

scaytrase:
@KingCrunch  @Ocramius 
Initially I has a workaround for this, which just drops the `nsSuffix` part if there is not `nsSeparator` used within. But I finished with the fact that it is useless for correct setups.

As I have already said the main idea is that you need to have two mapping file for a class to make broken setup work (one in the root directory and on in some subdirectory). Without this you'll get `throw MappingException::mappingFileNotFound($className, $filename);`, because `findMappingFile` is always looking for root-directory mapping file.

But I don't think that this is setup that was intended to use (or am I wrong?)

To keep support these setups I can submit a patch, if needed

mvrhov:
You have to submit at least one patch... As I said a couple of days ago.. The namespace separator used currently IS wrong.

scaytrase:
@mvrhov the namespace separtor used currently is the same as before. You can inspect the changes (and test files) to be sure it is still `.` (dot)

Two things changed:
1. relative path from mappings root to mappings files is not dropped anymore. 
2. namespace separator is configurable

mvrhov:
@scaytrase: Please look the at the FQCN again. In my post and the one made @KingCrunch this is probably true only on linux the / is used as a separator an not the \

KingCrunch:
Something I don't get: In PHP the namespace separator `\`. Why is it configureable at all? :confused:

scaytrase:
@KingCrunch we are talking about representation of namespace separator while converting namespace to filename. PHP's one is hardcoded to `\` as it should be.

KingCrunch:
@scaytrase OK, but I saw `Class 'FOS\UserBundle\Entity/model\User' does not exist`, so at least at one point the transformation to the actual classname failed.

scaytrase:
@KingCrunch I've look through FOSUserBundle doctrine configuration, and see no error there. Also nothing, that should be related with changes of this patch. `/model` part should not be passed to class name neither before, nor with this patch. I'll try to investigate the cause

scaytrase:
@KingCrunch Just a guess. If you temporary move mappings to another dir (like `Resources/config/mappings`) and change according mappings path at https://github.com/FriendsOfSymfony/FOSUserBundle/blob/master/FOSUserBundle.php
- it works?

```php
 $mappings = array(
            realpath(__DIR__ . '/Resources/config/mappings') => 'FOS\UserBundle\Model',
        );
```

lib/Doctrine/Common/Persistence/Mapping/Driver/SymfonyFileLocator.php
```diff
@@ -172,9 +186,17 @@ public function getAllClassNames($globalBasename = null)
...
-                        $classes[] = $this->prefixes[$path].'\\'.str_replace('.', '\\', $fileName);
+
+                        // Calculate namespace suffix for given prefix as a relative path from basepath to file path
+                        $nsSuffix = strtr(
+                            substr(realpath($file->getPath()), strlen(realpath($path))),
+                            $this->nsSeparator,
+                            '\\'
+                        );
+
+                        $classes[] = $this->prefixes[$path] . $nsSuffix . '\\' .str_replace($this->nsSeparator, '\\', $fileName);
```
eko: There is an issue here when Doctrine `auto_mapping` is enabled in Symfony and having a subdirectory like `Resources/config/doctrine/model` for instance.

eko:
I agree with @KingCrunch, I think an issue has been introduced here when having a subdirectory like `Resources/config/doctrine/model/` and Doctrine `auto_mapping` is set to true.

There are multiple bundles with a directory structure like this. We also use it in EkinoWordpressBundle: `https://github.com/ekino/EkinoWordpressBundle/tree/master/Resources/config/doctrine`

Can you take a look at this please @scaytrase?

I'm available if you need more feedback.

Thank you

scaytrase:
@eko @KingCrunch Yes, I think the problem is that `auto_mapping` looks for mappings at `Resources/config/doctrine/` instead of subdirectory (which started behave differently after this patch). The patch could be easily implemented, but I need a word of @Ocramius or @stof to be sure this is a correct setup, which should be supported. 

I've look through the docs and don't see any arguments, neither pro nor contra storing mapping files at the subdirectories. 

If they tell us, that subdirectories should be supported transparently (without any additional configuration), I'll submit a PR with fix.

eko:
@scaytrase Alright, thank you for considering this.

What do you think about supporting this transparently (as it was before this version?) @Ocramius @stof ?

Thank you all

eko:
Any update about that please?

Thank you

joskfg:
@scaytrase, Any progess?

victorfcm:
@scaytrase, Any progress?

PR #359, opened 2015-03-15
Fixed missing / incorrect docblocks

BenMorel:


lib/Doctrine/Common/Persistence/Mapping/Driver/PHPDriver.php
```diff
@@ -35,7 +35,7 @@
 class PHPDriver extends FileDriver
 {
     /**
-     * {@inheritDoc}
+     * @var ClassMetadata
      */
     protected $metadata;
```
BenMorel: Note: there is no `$metadata` in `FileDriver`, so `@inheritDoc` is invalid here.

Ocramius:
:+1:

PR #361, opened 2015-03-27
Fix static call in test

DavidPrevot:
Two tests were failing with the following error message:

Non-static method PHPUnit_Framework_MockObject_Generator::getMock()
should not be called statically, assuming $this from incompatible
context

### DIFF
--- a/tests/Doctrine/Tests/Common/Persistence/ManagerRegistryTest.php
+++ b/tests/Doctrine/Tests/Common/Persistence/ManagerRegistryTest.php
@@ -2,10 +2,10 @@
 
 namespace Doctrine\Tests\Common\Persistence;
 
-use \PHPUnit_Framework_MockObject_Generator;
 use Doctrine\Common\Persistence\AbstractManagerRegistry;
 use Doctrine\Tests\Common\Persistence\Mapping\TestClassMetadataFactory;
 use Doctrine\Tests\DoctrineTestCase;
+use PHPUnit_Framework_TestCase;
 
 /**
  * @groups DCOM-270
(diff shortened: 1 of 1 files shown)


tests/Doctrine/Tests/Common/Persistence/ManagerRegistryTest.php
```diff
@@ -64,12 +63,12 @@ public function testGetManagerForInvalidAliasedClass()
     }
 }
 
-class TestManager
+class TestManager extends DoctrineTestCase
```
Ocramius: Do we need to extend the `DoctrineTestCase`? Can't we just extend a `PHPUnit_Framework_TestCase`?

DavidPrevot: > Can't we just extend a PHPUnit_Framework_TestCase?

You’re right, PR updated, thanks.

Please note that the tests still fail when calling the file alone (within the global phpunit call it works fine):
```sh
$ phpunit tests/Doctrine/Tests/Common/Persistence/ManagerRegistryTest.php
[…]
There were 2 errors:

1) Doctrine\Tests\Common\Persistence\ManagerRegistryTest::testGetManagerForClass
ReflectionException: Class Doctrine\Tests\Common\Persistence\TestObject does not exist

/home/taffit/debian/php-doctrine-common/lib/Doctrine/Common/Persistence/AbstractManagerRegistry.php:197
/home/taffit/debian/php-doctrine-common/tests/Doctrine/Tests/Common/Persistence/ManagerRegistryTest.php:38
```

Ocramius:
@DavidPrevot thanks!

PR #366, opened 2015-04-28
Update DefaultFileLocator.php

marcel-burkhard:
It is implicit that a file extension needs a dot, thus it is unclear that we need to supply it in $fileExtension. I actually wasted quite some time trying to figure out what I'm doing wrong.


Ocramius:
:+1:

PR #367, opened 2015-05-08
Fix how namespace matching happens in SymfonyFileLocator

ptlis:
The problem was that namespace prefixes can overlap, and for classes
that would match more than one prefix only the first prefix would be
tested for. This meant that the findMappingFile and fileExists functions
would return different results depending on the order of the stored
prefixes.

For these methods to return the correct values the prefixes must be
listed in order of most to least specific. For example

    ['Vendor\Package\Entities', 'Vendor\Package\Entities\Blog']

would return the wrong result whereas

    ['Vendor\Package\Entities\Blog', 'Vendor\Package\Entities']

would return the correct one.

This changeset simply allows the class to be compared to more than one prefix, and updates one test as an exception message has changed.


ptlis:
Has anybody had a chance to look at this?

This is a very simple PR and it's very disheartening to see it languish for 4 months without any feedback either way.

Ocramius:
@ptlis didn't check so far. Thanks for the ping though.

What I see is a missing test here.

Ocramius:
Just re-checked this: it needs an additional test case for the scenario where multiple paths are mapped.

ptlis:
:+1: thanks @Ocramius , I'll try and find time to look at this over the weekend.

ptlis:
I've added tests to ensure that the prefix ordering no longer makes a difference, is there anything else I need to do to get this into a mergeable state?

Ocramius:
@ptlis merged, thanks!

Backported to 2.5.x in doctrine/common@e8768f6524336e747407a22ee4eb15e223dcdace
Backported to 2.6.x in doctrine/common@50b23081dc356c2cdd4eefb4ee2ca254c051cdc1

PR #369, opened 2015-05-22
[RFC] Load parent classes when loading metadata from cache

gonzalovilaseca:
This can be done more neatly, but before putting more time in it I'd like to check if it's a good idea.
I came across this solution because we had a problem with Gedmo extensions when caching metadata in Redis, this is our scenario:

```
Class A extends B;
```

`Class B` has a `Gedmo extension` mapped field.

On first request metadata for both `Class A` and `Class B` is cached in Redis.

On second request `AbstractClassMetadataFactory` loads `Class A` from Redis.

Gedmo extension loops through all `Class A` parents and calls method `hasMetadataFor($parentClass)` in `AbstractClassMetadataFactory`, only `Class A` metadata has been loaded previously, therefore `hasMetadataFor($parentClass)` returns false, and `Gedmo` then skips the parent class.

Should Gedmo extension load the metadata for the parent class or should it be done as proposed in this PR?

### DIFF
--- a/lib/Doctrine/Common/Persistence/Mapping/AbstractClassMetadataFactory.php
+++ b/lib/Doctrine/Common/Persistence/Mapping/AbstractClassMetadataFactory.php
@@ -212,6 +212,16 @@ abstract class AbstractClassMetadataFactory implements ClassMetadataFactory
                     $this->loadedMetadata[$realClassName] = $cached;
 
                     $this->wakeupReflection($cached, $this->getReflectionService());
+                    
+                    if ( ! $this->initialized) {
+                        $this->initialize();
+                    }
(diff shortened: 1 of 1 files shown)


Ocramius:
@gonzalovilaseca what is the exact problem that this is fixing? I understand that you want to change loading order, but what is the benefit?

gonzalovilaseca:
No, it's not the order. 
With this PR the parent classes are also pulled from cache.

gonzalovilaseca:
ping

gonzalovilaseca:
ping

Ocramius:
Needs test scenario that validates the use-case (failing before patch)

Majkl578:
Hello,
Thanks.

PR #373, opened 2015-07-10
composer: bump to PHPUnit 4.7

TomasVotruba:
PHPUnit 4 is here, 5 on the way

Min stability dev is not needed, ref: https://github.com/doctrine/common/commit/d2d2e0a64d4686dd0908c9226c8acfe5bcb4089f

### DIFF
--- a/tests/Doctrine/Tests/Common/ClassLoaderTest.php
+++ b/tests/Doctrine/Tests/Common/ClassLoaderTest.php
@@ -107,8 +107,8 @@ class ClassLoaderTest extends \Doctrine\Tests\DoctrineTestCase
 
     public function testSupportsTraitAutoloading()
     {
-        if (! function_exists('trait_exists')) {
-            $this->markTestSkipped('You need a PHP version that supports traits in order to run this test');
+        if (PHP_VERSION_ID < 50400) {
+            $this->markTestSkipped('Traits are only supported in PHP >=5.4.0');
         }
 
         $classLoader = new ClassLoader();
--- a/composer.json
+++ b/composer.json
@@ -22,7 +22,7 @@
     },
     "minimum-stability": "dev",
     "require-dev": {
-        "phpunit/phpunit": "~3.7"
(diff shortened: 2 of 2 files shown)


TomasVotruba:
This is probably caused by [minimum-stability:dev](https://github.com/doctrine/common/blob/ff72726b0876638fa7db2f28b0f26e3f1f6656ca/composer.json#L23), causing to install PhpUnit 4.8-dev which has some PHP 5.3 compatibility issues.

Why is this line needed anyway? Does this package depends on some unstable ones?

Ocramius:
@TomasVotruba can you check the failure btw? I think it's simply not skipping a test whereas it should

TomasVotruba:
@Ocramius Sure, will do in few hours.

stof:
the failure is related to the fact that some trait-based tests are loaded on PHP 5.3 while they should not

TomasVotruba:
@stof Have you came across it already? If so, what was the solution?

stof:
Well, ``ClassLoaderTests::testSupportsTraitAutoloading`` does not seem to be skipped properly

TomasVotruba:
Shall be fixed.

tests/Doctrine/Tests/Common/ClassLoaderTest.php
```diff
@@ -107,8 +107,8 @@ public function testSupportsInterfaceAutoloading()
 
     public function testSupportsTraitAutoloading()
     {
-        if (! function_exists('trait_exists')) {
```
Ocramius: Sneaky polyfills? :-(

TomasVotruba: [Probably](https://github.com/TomasVotruba/common/commit/9baf712e9439a14b9a1cec4b310401c58d1fb720#diff-e6bdb06863ccfe13a910a0485b3a74faR99) :/

TomasVotruba:
Thanks! @stof & @Ocramius

PR #374, opened 2015-07-21
getManagerForClass should try to retrieve the default manager

elernonelma:
getManagerForClass should try to retrieve the default manager instead of the first one found in the list.

### DIFF
--- a/lib/Doctrine/Common/Persistence/AbstractManagerRegistry.php
+++ b/lib/Doctrine/Common/Persistence/AbstractManagerRegistry.php
@@ -200,9 +200,11 @@ abstract class AbstractManagerRegistry implements ManagerRegistry
         }
 
         // First, try to retrieve the default one
-        $defaultManager = $this->getService($this->managers[$this->getDefaultManagerName()]);
-        if (!$defaultManager->getMetadataFactory()->isTransient($class)) {
-            return $defaultManager;
+        if (isset($this->managers[$this->getDefaultManagerName()])) {
+            $defaultManager = $this->getService($this->managers[$this->getDefaultManagerName()]);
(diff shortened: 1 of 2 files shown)


TomasVotruba:
Could you also provide tests?

elernonelma:
The comprehension of the actual test was painful. Not great here (the test itself doesn't test a lot), but it covers the non regression of existing test. If you think something can be done better easily, I will try it but without a refactoring of the actual test, I think it will be hard.

elernonelma:
To abort the context of use, I have N managers for my entity, and I choose a default one. I supposed then that this method would return the default one, and not the first one declared. It may have a reason to this?

BobWeb:
I encounter the same problem as elernonelma. Do you have information about the pull request's validation, because it would really help !

Ocramius:
> I have N managers for my entity

This assumption seems wrong: the registry should map each entity to a separate manager.

elernonelma:
How can it be wrong? My entity must be used by N different databases, handled by N manager.

Ocramius:
Yes, but this class assumes that only one manager is used per type of class. The idea is to bridge different domains together via different ORM/ODM instances.

elernonelma:
Yes, it is exactly how it works and exactly what i expect it to. But I would expect that it would try to use the default manager first, instead of the first one found in the declaration.

Ocramius:
The "default" one is supposed to work as a fallback

elernonelma:
I dont expected it to work as a fallback, as it's default I expect it to works as a default. Did not find any information about it. Where did you see that?
Here, it does not works as a fallback.

Majkl578:
Hello,
Thanks.

PR #375, opened 2015-07-24
Allow access to variables in concrete implementations

mbbender:
It seems the variables in this abstract class should be protected instead of private so concrete implementations can modify them as needed.


TomasVotruba:
You can access properties with getters, like `$this->getName()`. This is to protect properties against change. Having protected would allow such a change.

Ocramius:
This sort of change requires stronger use-cases in order to be accepted. Closing, as the use-case is missing.

PR #380, opened 2015-09-03
Add .gitattributes (fixes #379)

### LINKED ISSUE #379: Exclude tests from dist
I'm using doctrine in a symfony project and get it via composer with ```--prefer-dist``` option. This type of installation also grabs the tests in ```doctrine/common```, which causes eclipse PHP builder to flag an error:

screen shot 2015-09-03 at 13 12 13

_[VariadicTypeHintClass.php](https://github.com/doctrine/common/blob/master/tests/Doctrine/Tests/Common/Proxy/VariadicTypeHintClass.php) has got some PHP5.6-specific syntax, while the whole repository seems to be still compatible with 5.3 (according to [composer.json](https://github.com/doctrine/common/blob/master/composer.json))._

Sometimes a special ```.gitattributes``` file is added to the project root dir, which forces github to exclude certain folders from the zip archives that are being fetched by composer. This is done [in Doctrine2](https://github.com/doctrine/doctrine2/blob/master/.gitattributes), for example.

Could the same practice be applied in ``doctrine/common``?

kachkaev:


.gitattributes
```diff
@@ -0,0 +1,8 @@
+/tests export-ignore
```
Ocramius: Should not have the slash at the beginning

kachkaev: Not sure. Same stuff in doctrine/doctrine2:
https://github.com/doctrine/doctrine2/blob/master/.gitattributes

kachkaev: I only removed lines that are not relevant to doctrine/common

stof: well, it actually make sense. We want to exclude the tests folder at the root of the repo, but not folders named tests which could appear elsewhere (for code using a ``Tests`` namespace to provide test helpers for libraries for instance)

kachkaev: @stof, there's a similar issue with symfony/symfony itself by the way. I just noticed that there are a few deliberately broken xmls in */Tests/ (e.g. [invalid.xml in Config component](https://github.com/symfony/symfony/blob/2.8/src/Symfony/Component/Config/Tests/Fixtures/Util/invalid.xml)).

In that repository it might be more reasonable to also add .gitattributes, but remove slash before "tests". But you're right that there is a potential problem with namespaces.

stof: @kachkaev Symfony explicitly reverted the exclusion of tests of the archive a while ago

Ocramius:
Fixed on merge

PR #383, opened 2015-09-18
Silence chmod() warnings

mpdude:
This may be required on some filesystems, most notably in some situations when exporting via NFS into a Vagrant box. Haven't dug into details but seems to be an issue if the owner of the file is not the user trying to `chmod` (?).


mpdude:
doctrine/annotations uses as silenced `@chmod` call as well: https://github.com/doctrine/annotations/commit/f25c8aab83e0c3e976fd7d19875f198ccf2f7535

Ocramius:
I just re-applied these changes manually to `master` and `2.5`, thanks!

PR #387, opened 2015-10-21
Fixed fatal error in AbstractManagerRegistry

enumag:
```
AbstractManagerRegistry::getManagerForClass('Doctrine\Common\Persistence\Proxy');
// Fatal Error Call to a member function getName() on boolean
```

### DIFF
--- a/lib/Doctrine/Common/Persistence/AbstractManagerRegistry.php
+++ b/lib/Doctrine/Common/Persistence/AbstractManagerRegistry.php
@@ -196,6 +196,9 @@ abstract class AbstractManagerRegistry implements ManagerRegistry
 
         $proxyClass = new \ReflectionClass($class);
         if ($proxyClass->implementsInterface($this->proxyInterfaceName)) {
+            if (!$proxyClass->getParentClass()) {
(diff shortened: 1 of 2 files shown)


Ocramius:
@enumag we need a test to validate this change: why is one of the proxies not extending a base class btw?

enumag:
@Ocramius This is a sideeffect of using ManagerRegistry in my implementation of `Symfony\Component\Validator\Mapping\Loader\LoaderInterface`. Basically I wanted to add some rules for all of my Doctrine entities. The loadClassMetadata method is called not only for the class itself but also for all its parents and interfaces.

I'll try to add a test.

enumag:
@Ocramius Test added.

Ocramius:
Thanks!

Ocramius:
Merged, thanks!

`master`: doctrine/common@79fdef35a79a3694e5ae552575d23fd7dc8482c2
`2.5.x`: doctrine/common@1fb4d85982a4c18bff80e05659b110312f416de7

TomasVotruba:
Awesome!

PR #390, opened 2015-10-24
composer: drop duplicated "archive"

TomasVotruba:
This is already managed in https://github.com/doctrine/common/blob/master/.gitattributes


Ocramius:
@TomasVotruba thanks!

TomasVotruba:
:+1:

PR #376, opened 2015-08-26
Proxy: Added support for PHP 7 return & parameter type hints

Majkl578:
This PR adds support for PHP 7 return types and parameter type hints on generated proxies.

PR #391, opened 2015-11-18
PHP7: Proxy generator to handle return type

wivern-co-uk:
Hi, it's initial PR to raise a discussion and make ProxyGenerator works with php7 return types

### DIFF
--- a/lib/Doctrine/Common/Proxy/ProxyGenerator.php
+++ b/lib/Doctrine/Common/Proxy/ProxyGenerator.php
@@ -950,6 +950,12 @@ EOT;
         if (method_exists($parameter, 'isCallable') && $parameter->isCallable()) {
             return 'callable';
         }
+        
+        $parameterType = (string) $parameter->getType();
+        
+        if (in_array($parameterType, ['int', 'bool', 'float', 'string'])) {
+            return $parameterType;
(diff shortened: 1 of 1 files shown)


Ocramius:
@wivern-co-uk much appreciated, but we'd still require test cases to validate this logic :-)

wivern-co-uk:
Looks like it's duplicated as of https://github.com/doctrine/common/pull/376

lib/Doctrine/Common/Proxy/ProxyGenerator.php
```diff
@@ -802,6 +807,23 @@ private function generateMethods(ClassMetadata $class)
...
+     *
+     * @param ReflectionMethod $method
+     *
+     * @return string
+     */
+    protected function getReturnType(\ReflectionMethod $method){
+        $returnType = $method->getReturnType();
+
+        if (! $returnType->isBuiltin()) {
+            $returnType = '\\' . $returnType;
```
stof: this will not handle ``self`` properly, will it ?

Majkl578:
This is indeed duplicate of #376.

Ocramius:
Dupe of #376 indeed. Closing

PR #394, opened 2015-12-04
Improve MappingData retrieval by caching Results.

gries:
In cases where mapping Data is not cached by calling ```initilaize();```
It makes sense to cache results after parsing the Yaml file.


Ocramius:
@gries do you think that you can test the change? Needs a test subclassing the file driver

gries:
I have added a test that calls the ```getElement()```  twice and makes sure that the Yaml file is only loaded once.

gries:
any update on this?

Ocramius:
@gries merged and ready for 2.7.0

PR #701, opened 2015-12-07
Allow Traits to be mapped with Yaml

robhunt3r:
I think this is not a real patch because it may mess with something else, but I get it to work in my project.

The problems I've found are related to Gedmo's library with, for example, Sluggable and Timestampable, which doesn't trigger and set them to null, but maybe this is a start.

Why this PR? Because in a lot of projects people use YAML to get a clean code, but sometimes we need to use Traits, and Doctrine doesn't like mixing Annotations and Yaml, so our traits are being ignored, and also, there's no YAML mapping for Traits (see http://stackoverflow.com/questions/34133014/doctrine2-mixing-yaml-entities-with-annotation-traits-doesnt-work)

The solution without this PR is to create your Trait and remap it in every Entity that uses it, which is precisely the aim of the Traits, use a model in several places without redeclaring them...

How can you try?

Simple create two files, one is an Entity and the other one is a Trait and map them this way
```YAML
Foo:
    type: entity
    id:
        id:
            type: integer
            generator:
                strategy: AUTO
Bar:
    type: mappedSuperclass
        name:
            type: string
            length: 255
        slug:
            type: string
            length: 255
```
```PHP
class Foo
{
    use Bar;

    private $id;

    /**
     * @return mixed
     */
    public function getId()
    {
        return $this->id;
    }
}

trait Bar
{
    protected $name;
    protected $slug;

    /**
     * @return mixed
     */
    public function getName()
    {
        return $this->name;
    }

    /**
     * @param mixed $name
     */
    public function setName($name)
    {
        $this->name = $name;
    }

    /**
     * @return mixed
     */
    public function getSlug()
    {
        return $this->slug;
    }

    /**
     * @param mixed $slug
     */
    public function setSlug($slug)
    {
        $this->slug = $slug;
    }
}
```

I repeat, this is not a real PR because I didn't have the time to test everything, it's just a start, it just checks for a Class or a Trait and merge every property in order to generate a corrrect hierarchy.

### DIFF
--- a/lib/Doctrine/Common/Persistence/Mapping/RuntimeReflectionService.php
+++ b/lib/Doctrine/Common/Persistence/Mapping/RuntimeReflectionService.php
@@ -37,11 +37,11 @@ class RuntimeReflectionService implements ReflectionService
      */
     public function getParentClasses($class)
     {
-        if ( ! class_exists($class)) {
+        if ( ! class_exists($class) && ! trait_exists($class)) {
             throw MappingException::nonExistingClass($class);
         }
 
-        return class_parents($class);
+        return array_merge(class_uses($class), class_parents($class));
     }
 
     /**


Ocramius:
This patch is invalid, because the semantics of `getParentClasses` should return reflections of classes, not interfaces or traits.

Additionally, traits are quite complicated to deal with, because they don't really exist as a concept at runtime, and therefore should really be excluded.

The only current implementation that supports traits mapping is the annotations system, because annotations exist on the symbol they are defined on (property/method), and not on the trait "concept".

If you want to support traits, then additional API should be created (and consumed) by the mapping drivers directly, as far as I can tell at first glance. From my perspective, this sort of functionality is to be avoided anyway (re-map the fields at entity level instead).

Closing this PR as invalid for now.

Issue #327, opened 2014-07-14
rename error in ProxyGenerator


Draeli:
Hello,

Actually, I have this hapening, I don't have idea why, I give you my backtrace, this not happen each time I refresh page but some time.

``` php
( ! ) Warning: rename(/app/cache/dev/doctrine/orm/Proxies\__CG__MyCommonBundleEntityVisibilityInformation.php.53c3cccd0a5415.53070441,/app/cache/dev/doctrine/orm/Proxies\__CG__MyCommonBundleEntityVisibilityInformation.php): in \vendor\doctrine\common\lib\Doctrine\Common\Proxy\ProxyGenerator.php on line 305
```

backtrace :

```
Call Stack
#   Time    Memory  Function    Location
18  0.6984  11450624    Doctrine\ORM\EntityRepository->findOneBy( ) ..\UserManager.php:68
19  0.7052  11784056    Doctrine\ORM\Persisters\BasicEntityPersister->load( )   ..\EntityRepository.php:196
20  0.7709  12390192    Doctrine\ORM\Internal\Hydration\AbstractHydrator->hydrateAll( ) ..\BasicEntityPersister.php:756
21  0.7709  12390280    Doctrine\ORM\Internal\Hydration\SimpleObjectHydrator->hydrateAllData( ) ..\AbstractHydrator.php:140
22  0.7710  12395136    Doctrine\ORM\Internal\Hydration\SimpleObjectHydrator->hydrateRowData( ) ..\SimpleObjectHydrator.php:48
23  0.7715  12411464    Doctrine\ORM\UnitOfWork->createEntity( )    ..\SimpleObjectHydrator.php:138
24  0.7869  12653744    Doctrine\Common\Proxy\AbstractProxyFactory->getProxy( ) ..\UnitOfWork.php:2676
25  0.7869  12653832    Doctrine\Common\Proxy\AbstractProxyFactory->getProxyDefinition( )   ..\AbstractProxyFactory.php:119
26  0.7882  12666032    Doctrine\Common\Proxy\ProxyGenerator->generateProxyClass( ) ..\AbstractProxyFactory.php:218
27  0.7930  12695936    rename ( )  ..\ProxyGenerator.php:305
```

brunowego:
+1

diegopso:
+1

chielsen:
+1

ragboyjr:
:+1:

Apezoumon123:
somebody help me i have the same problem

[...]

PR #702, opened 2015-12-12
Add strategy autogenerate the proxy class when the proxy file does no…

### LINKED ISSUE #362: Warning with php rename function, in ProxyGenerator class
Hello,

Sometimes i got a warning with the `rename` function, called in `Doctrine\Common\Proxy\ProxyGenerator:305`

```
Warning: rename(C:\Windows\TEMP\__CG__AppEntitiesCategory.php.551eee3a298629.89032411,C:\Windows\TEMP\__CG__AppEntitiesCategory.php)
```

I got this warning when I send multiple xhr request, so i guess there is a conflict with temporaries files ?

I temporarily fixed this by adding a `@` before the function call to disable error, but i hope there's a better way to fix this

Note: I got the same warning when I configure a specific cache directory

```php
// this directory exists and is writeable
Setup::createConfiguration($isDevMode, __DIR__.'/../../cache/doctrine');
```

Koc:
…t exist or outdated.

This should improve DX a lot. Fixes #362, #327

### DIFF
--- a/lib/Doctrine/Common/Proxy/AbstractProxyFactory.php
+++ b/lib/Doctrine/Common/Proxy/AbstractProxyFactory.php
@@ -60,6 +60,15 @@ abstract class AbstractProxyFactory
     const AUTOGENERATE_FILE_NOT_EXISTS = 2;
 
     /**
+     * Autogenerate the proxy class when the proxy file does not exist or outdated.
+     *
+     * This strategy causes a file exists and filemtime call.
+     *
+     * @var integer
+     */
+    const AUTOGENERATE_FILE_OUTDATED = 4;
+
+    /**
      * Generate the proxy classes using eval().
      *
      * This strategy is only sane for development, and even then it gives me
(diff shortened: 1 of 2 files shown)


lib/Doctrine/Common/Proxy/AbstractProxyFactory.php
```diff
@@ -216,6 +226,13 @@ private function getProxyDefinition($className)
                     require $fileName;
                     break;
 
+                case self::AUTOGENERATE_FILE_OUTDATED:
+                    if ( ! file_exists($fileName) || filemtime($classMetadata->getReflectionClass()->getFileName()) > filemtime($fileName)) {
```
Ocramius: what if the original class file name doesn't exist?

Koc: I haven't know that proxy generation can be used for internal classes. In this case we shouldn't generate new proxy class. What do you think?

Ocramius: It should probably re-generate it

johndodev:
+1

Koc:
@Ocramius I've updated the PR. IMHO there is no sense autogenerate proxy class for internall classes on each request because they are changes only when PHP/extension version changes, but in this case cache should be flushed anyway.

Ocramius:
Checking this, it's not a sufficient change. Specifically, we'd need to track all ancestors in a way similar to https://github.com/doctrine/annotations/pull/105

A simpler approach would be to use something like https://github.com/Ocramius/PackageVersions to simply lock dependencies at package level or at entire project level, but it would require us to code-sign our proxies.

Koc:
good catch, thanx for giving right direction

mikeSimonson:
@Koc Should I close this PR and https://github.com/doctrine/DoctrineBundle/pull/636 then ?

Koc:
Please no. I will continue on it. This problem occurs very ofthen on concurrent requests in dev env

Majkl578:
Hello, I'm going to close this PR as we will not be maintaining Doctrine\Common\Proxy anymore. As an alternative, you can use [Ocramius/ProxyManager. ](https://github.com/ocramius/proxymanager). Thank you for the PR.
If you think this PR is absolutely must, let me know, I'll reopen and we can decide.

PR #719, opened 2016-05-25
Improve the phpdoc of ObjectRepository

mnapoli:
`find()` and `findOne()` methods can return `null` but it wasn't documented. I had to go to the implementation to be absolutely certain.


Ocramius:
@mnapoli thanks!

PR #720, opened 2016-06-06
Fixed detection of identifier getter with PHP 7 return type

janlanger:


lib/Doctrine/Common/Proxy/ProxyGenerator.php
```diff
@@ -37,7 +37,7 @@ class ProxyGenerator
      * Used to match very simple id methods that don't need
      * to be decorated since the identifier is known.
      */
-    const PATTERN_MATCH_ID_METHOD = '((public\s+)?(function\s+%s\s*\(\)\s*)\s*{\s*return\s*\$this->%s;\s*})i';
+    const PATTERN_MATCH_ID_METHOD = '((public\s+)?(function\s+%s\s*\(\)\s*)\s*(?::\s*[\w\\\\]+\s*)?{\s*return\s*\$this->%s;\s*})i';
```
Ocramius: php type identifiers follow the regex `'/^[a-zA-Z_\x7f-\xff][a-zA-Z0-9_\x7f-\xff]+$/'`

Ocramius: Also: needs a test case

janlanger: @Ocramius But there can be more identifiers separated by backslash - namespaces. This version is simple, I could extend it to include `\x7f-\xff` or use strict version like `(?::\s*\\?[a-z_\x7f-\xff][\w\x7f-\xff]*(?:\\[a-z_\x7f-\xff][\w\x7f-\xff]*)*\s*)`. But is strict really necessary? The checked code have to be valid PHP anyway and there cannot be any code between closing `)` and opening `{`

I'll try to prepare test, is there any existing testcase for this feature? I didn't find any...

Ocramius: @janlanger well, general purpose lib => general purpose solution :-)

In think tests for this particular scenario are somewhere in https://github.com/doctrine/common/blob/faf7d81273eeffb84d524006daf85d37a14ac42a/tests/Doctrine/Tests/Common/Proxy/ProxyClassGeneratorTest.php

janlanger:
@Ocramius updated

lib/Doctrine/Common/Proxy/ProxyGenerator.php
```diff
@@ -37,7 +37,7 @@ class ProxyGenerator
      * Used to match very simple id methods that don't need
      * to be decorated since the identifier is known.
      */
-    const PATTERN_MATCH_ID_METHOD = '((public\s+)?(function\s+%s\s*\(\)\s*)\s*{\s*return\s*\$this->%s;\s*})i';
+    const PATTERN_MATCH_ID_METHOD = '((public\s+)?(function\s+%s\s*\(\)\s*)\s*(?::\s*\\\\?[a-z_\x7f-\xff][\w\x7f-\xff]*(?:\\\\[a-z_\x7f-\xff][\w\x7f-\xff]*)*\s*)?{\s*return\s*\$this->%s;\s*})i';
```
Ener-Getick: what about using ``[^\{]*`` instead ?

tests/Doctrine/Tests/Common/Proxy/ProxyLogicTypehintsTest.php
```diff
@@ -0,0 +1,199 @@
...
+use Doctrine\Common\Proxy\ProxyGenerator;
+use Doctrine\Common\Proxy\Proxy;
+use Doctrine\Common\Proxy\Exception\UnexpectedValueException;
+use Doctrine\Common\Persistence\Mapping\ClassMetadata;
+use PHPUnit_Framework_TestCase;
+
+/**
+ * Test that identifier getter does not cause lazy loading. These tests make assumptions about the structure of LazyLoadableObjectWithTypehints
+ *
+ * @author Marco Pivetta <ocramius@gmail.com>
```
Ener-Getick: should be your own name ;)

lib/Doctrine/Common/Proxy/ProxyGenerator.php
janlanger: @Ener-Getick That would not be "general purpose solution" :) See https://github.com/doctrine/common/pull/720#discussion_r65870837 and following,

Ener-Getick: I would prefer simplicity over strictness in this case but looks good anyway ☺

Ocramius: Correctness is required for generic-purpose libraries, or else there will be dragons.

tests/Doctrine/Tests/Common/Proxy/ProxyLogicTypehintsTest.php
janlanger: The code is basically modified copy of ProxyLogicTest. I added myself as second author :)

janlanger:
Updated & rebased on master

Ocramius:
@janlanger merging, thanks!

solcik:
Hello,
as this pull request accepted option to add a return type for identifier getter method.
Is it possible to accept typecasting in regex **PATTERN_MATCH_ID_METHOD**?

Getter like this one, will generate proxy below.

```php
public function getId(): int
{
    return $this->id;
}
```

```php
/**
 * {@inheritDoc}
 */
public function getId(): int
{
    if ($this->__isInitialized__ === false) {
        return (int)  parent::getId();
    }


    $this->__initializer__ && $this->__initializer__->__invoke($this, 'getId', []);

    return parent::getId();
}
```

So when I call **::getId()** in not initialized entity, it throws **TypeError**, because **$this->id** is *string*.

```
TypeError
Return value of ...::getId() must be of the type integer, string returned
```

So It would be really nice, if I could typecast identifier and not end up with

```php
public function getId(): int
{
    return (int) $this->id;
}
```

```php
/**
 * {@inheritDoc}
 */
public function getId(): int
{
    $this->__initializer__ && $this->__initializer__->__invoke($this, 'getId', []);

    return parent::getId();
}
```

Ocramius:
David, that sounds super-broken. The ORM already casts identifiers to their
mapped types: casting yourself too indicates a design problem in your
codebase first, and the ORM shouldn't help you in supporting bad practices.

PR #721, opened 2016-06-11
small improvements from humbug

BreiteSeite:
Hi,

i tried humbug for the first time (against your codebase) and wanted to share some slight modifications i made to fix 2 "escapes".

I would take no offence if you didn't want to merge such a small change.


.gitignore
```diff
@@ -8,3 +8,4 @@ vendor/
 composer.lock
 doctrine-common-*.tar
 doctrine-common-*.tar.gz
+humbuglog.txt
```
stof: we have an ignored ``reports`` folder for tool reports. Please put the file in it instead of ignoring a new root file

lib/Doctrine/Common/Proxy/Autoloader.php
```diff
@@ -49,7 +49,11 @@ public static function resolveFile($proxyDir, $proxyNamespace, $className)
             throw InvalidArgumentException::notProxyClass($className, $proxyNamespace);
         }
 
-        $className = str_replace('\\', '', substr($className, strlen($proxyNamespace) + 1));
+        // remove proxy namespace from class name
+        $className = substr($className, strlen($proxyNamespace));
```
stof: why remove this `` + 1`` ?

BreiteSeite: Doesn't seem to make a difference. Humbug mutated this one to -1 i think and test were still green. I even added another dataset for the test to be sure.

.gitignore
BreiteSeite: Thanks, i will update accordingly.

humbug.json.dist
```diff
@@ -0,0 +1,11 @@
...
+    "source": {
+        "directories": [
+            "lib\/Doctrine"
+        ]
+    },
+    "timeout": 10,
+    "logs": {
+        "text": "reports/humbuglog.txt"
+    }
+}
```
Ocramius: EOF EOL still needed

lib/Doctrine/Common/Proxy/Autoloader.php
Ocramius: This is the difference: https://3v4l.org/tkrdl

Copying code for reference:

```php
$className = 'Foo\\Bar\\Baz';
$proxyNamespace = 'Foo';

var_dump(str_replace('\\', '', substr($className, strlen($proxyNamespace) + 1)));
var_dump(substr($className, strlen($proxyNamespace)));
```

lib/Doctrine/Common/Proxy/Autoloader.php
```diff
@@ -49,7 +49,11 @@ public static function resolveFile($proxyDir, $proxyNamespace, $className)
...
+
+        // remove namespace separators from remaining class name
+        $className = str_replace('\\', '', $className);
```
Ocramius: Don't overwrite the variable if its semantic meaning changes

lib/Doctrine/Common/Proxy/Autoloader.php
BreiteSeite: That comparison is a bit off, as the first var_dump includes the str_replace, while the second does not. We probably should compare this: https://3v4l.org/2vgog

Also, i'm not sure how relevant that is, as the slashes are removed with the str_replace in the next line anways.

Ocramius: Ah, yep, that is indeed true. Indeed irrelevant, as doctrine always puts a `\\` between the proxy namespace and the class name.

Thanks, indeed flawed example by me

lib/Doctrine/Common/Proxy/Autoloader.php
BreiteSeite: I totally agree with you here. Do you think the new names are more fitting?

Ocramius:
@BreiteSeite thanks!

PR #722, opened 2016-06-14
Allows invalid type hints

Ener-Getick:
The doctrine proxies can't be generated when a class contains an invalid type hint but this behaviour causes issues in case a type hint is voluntary invalid (see https://github.com/FriendsOfSymfony/FOSCommentBundle/issues/538#issuecomment-225862002).

So my proposal is to simply ignore methods containing invalid type hints because they won't be callable anyway.

### DIFF
--- a/lib/Doctrine/Common/Proxy/ProxyGenerator.php
+++ b/lib/Doctrine/Common/Proxy/ProxyGenerator.php
@@ -767,6 +767,14 @@ EOT;
                 continue;
             }
 
+            // Checks earlier that the parameters type hint are valid.
+            try {
+                $parameters = $this->buildParametersString($class, $method, $method->getParameters());
+            } catch (UnexpectedValueException $exception) {
+                // in case a type hint is invalid, the method won't be callable anyway
+                continue;
+            }
+
             $methodNames[$name] = true;
             $methods .= "\n    /**\n"
                 . "     * {@inheritDoc}\n"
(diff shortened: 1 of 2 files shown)


Ener-Getick:
It's also possible to use [this solution](https://secure.php.net/manual/en/reflectionparameter.getclass.php#108620) in case we really want to generate the proxified method.

Ocramius:
We can't fix this, as reflection would simply crash when an invalid type hint is provided. If the type hint is not there, we can't do much about it.

Closing as `can't fix` and `invalid`, as the bug is clearly in the source entity.

Ener-Getick:
@Ocramius this PR proposes to not generate a method if reflection crashes.
This allows a behavior allowed by php and won't cause cases where the proxified class wouldn't be correctly loaded because the classes not generated won't be callable.

Ocramius:
> @Ocramius this PR proposes to not generate a method if reflection crashes.

not gonna happen, that would be an edge-case scenario leading to more "WTF" situation than what it solves.

stof:
We could fix this for PHP 7+ by using ``$parameter->getType`` even for non-builtin types, as it will support this case properly (it won't try to build a ReflectionClass for the typehinted class)

Ener-Getick:
Indeed @stof proposal looks much better. @Ocramius would you accept such PR

Ocramius:
Can do, yes

PR #723, opened 2016-06-14
Add enum abstraction

Deltachaos:
Adding abstraction to handle enum types.

For working with enum objects in doctrine and becuase of the fact that there is no standard way of defining enums currently we need to abstract the handling for the dbal in a way that it can be extended to support other enum libraries and maybe a native enum implementation in the future.

Tests will be added as soon as people show there interest in the feature.

### DIFF
--- a/bootstrap.php
+++ b/bootstrap.php
@@ -1,0 +1,4 @@
+<?php
+
+\Doctrine\Common\Enum\Factory::registerEngine(new \Doctrine\Common\Enum\Engine\SPL());
+\Doctrine\Common\Enum\Factory::registerEngine(new \Doctrine\Common\Enum\Engine\MyCLabs());
--- a/composer.json
+++ b/composer.json
@@ -26,7 +26,8 @@
     "autoload": {
         "psr-4": {
             "Doctrine\\Common\\": "lib/Doctrine/Common"
-        }
+        },
+        "files": [ "bootstrap.php" ]
     },
     "extra": {
         "branch-alias": {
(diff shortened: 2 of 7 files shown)


bootstrap.php
```diff
@@ -0,0 +1,4 @@
+<?php
+
+\Doctrine\Common\Enum\Factory::registerEngine(new \Doctrine\Common\Enum\Engine\SPL());
+\Doctrine\Common\Enum\Factory::registerEngine(new \Doctrine\Common\Enum\Engine\MyCLabs());
```
stof: a total -1 for a global state being initialized eagerly for all projects using the library

stof:
Idon't think this should be added in the doctrine/common repo. the trend is to have more dedicated libraries rather than a big common one

deeky666:
I agree with @stof here. Also there are chances we'll get native enums sooner or later natively as of [this RFC](https://wiki.php.net/rfc/enum). Currently I don't see any reason to add this to Doctrine. As of DBAL, we cannot add an abstracted Enum type anyways as DBAL's type system is static and Enums are not. There have been a number of discussions about this in the past. [Please also refer to the documentation](http://doctrine-orm.readthedocs.io/projects/doctrine-orm/en/latest/cookbook/mysql-enums.html).

Ocramius:
Closing as per @deeky666's feedback, with which I agree strongly. Without a strongly-typed userland implementation of ENUMs (RFC above, for example), we are just having the usual old "stringly typed" enums that are likely buggy and problematic anyway.

PR #725, opened 2016-06-19
No reason to increment this

gadelat:


Ocramius:
@gadelat could you clarify on the change? If I understand this correctly, doctrine always dumped a maximum of 3 levels by default so far, while this change leads to dumping 2 levels again?

Ocramius:
Oh, wait, I'm confusing `$i++` and `++$i` semantics. :facepalm:

Alright, merging, thanks :-)

PR #726, opened 2016-06-22
ManagerRegistry::resetManager() does not return the new manager

ogizanagi:
As discussed in https://github.com/doctrine/common/pull/65/files#r68011118, the interface's docblock is confusing, as the implementations never returned the new instance of the named manager, but instead expects the next call to `getManager()` to do so.

Thus, the docblock should be changed accordingly.

### DIFF
--- a/lib/Doctrine/Common/Persistence/AbstractManagerRegistry.php
+++ b/lib/Doctrine/Common/Persistence/AbstractManagerRegistry.php
@@ -220,5 +220,7 @@ abstract class AbstractManagerRegistry implements ManagerRegistry
         // force the creation of a new document manager
         // if the current one is closed
         $this->resetService($this->managers[$name]);
+
+        return $this->getManager($name);
     }
 }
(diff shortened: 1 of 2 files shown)


Ocramius:
@ogizanagi looks like the implementation is wrong here, then. If the interface specified this, then somebody may already have implemented it correctly, and relied on it. Implementation should be changed instead, IMO

ogizanagi:
@Ocramius : Thanks for your opinion on this. Of course you're right, even if I think it's most unlikely someone implemented the interface, without using the `AbstractManageRegistry` or even needing to override it. But still, it is possible.

I've update the PR to change the `AbstractManageRegistry::resetManager()` implementation.

Ocramius:
@ogizanagi can you provide a test case as well?

stof:
This forces to reinstantiate immediately instead of waiting for the next time needing it, so this might be less optimal.

I would be in favor of updating the phpdoc to match the actual behavior of the code (the code works the same since years)

Ocramius:
You don't know all implementations

Ocramius:
@ogizanagi argh, I just noticed that. This fix would only land in 2.6.2 though. Can you recreate the PR onto `master`? We'd then backport on merge.

ogizanagi:
@Ocramius : Done in #727

Ocramius:
@ogizanagi thanks! Closing here.

PR #728, opened 2016-07-18
Detect id getter methods named exactly as id field

OJezu:
Support not loading full entity when calling method named like id field
(e.g. `id()`) on Proxy.

Some people are using naming convention that does not prefix getters with "get" and uses instead just the field name on (usually very rare) getter function.
Doctrine tries to detect id getter methods and avoid loading proxied entities when such method is called, but it does not support getters named exactly the same as the underlying id field. This PR adds support for such methods to already existing "get" prefixed id getter method detection.

### DIFF
--- a/lib/Doctrine/Common/Proxy/ProxyGenerator.php
+++ b/lib/Doctrine/Common/Proxy/ProxyGenerator.php
@@ -838,12 +838,15 @@ EOT;
      */
     private function isShortIdentifierGetter($method, ClassMetadata $class)
     {
-        $identifier = lcfirst(substr($method->getName(), 3));
+        $identifier = (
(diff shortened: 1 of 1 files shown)


Ocramius:
Requires test case. Look at https://github.com/doctrine/common/pull/720 for reference

Majkl578:
Hello, I'm going to close this incomplete PR as we will not be maintaining Doctrine\Common\Proxy anymore. As an alternative, you can use [Ocramius/ProxyManager. ](https://github.com/ocramius/proxymanager). Thank you for the PR.

Ocramius:
Note: this behaviour already works out of the box with the 3.x implementation

PR #729, opened 2016-07-20
Util\Debug: export parent class private attributes and DateTimeImmutable

Slamdunk:
This patch enhances DateTime export for the new DateTimeImmutable and fixes the behaviour dumping also the private attributes of parent class emulating `print_r`

### DIFF
--- a/lib/Doctrine/Common/Util/Debug.php
+++ b/lib/Doctrine/Common/Util/Debug.php
@@ -109,7 +109,7 @@ final class Debug
                 }
             } else if ($isObj) {
                 $return = new \stdclass();
-                if ($var instanceof \DateTime || $var instanceof \DateTimeInterface) {
+                if ($var instanceof \DateTimeInterface) {
                     $return->__CLASS__ = get_class($var);
                     $return->date = $var->format('c');
                     $return->timezone = $var->getTimezone()->getName();
--- a/lib/Doctrine/Common/Util/Debug.php
+++ b/lib/Doctrine/Common/Util/Debug.php
@@ -114,7 +114,6 @@ final class Debug
                     $return->date = $var->format('c');
                     $return->timezone = $var->getTimezone()->getName();
                 } else {
-                    $reflClass = ClassUtils::newReflectionObject($var);
                     $return->__CLASS__ = ClassUtils::getClass($var);
 
                     if ($var instanceof Proxy) {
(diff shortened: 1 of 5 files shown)


lib/Doctrine/Common/Util/Debug.php
```diff
@@ -109,10 +109,10 @@ public static function export($var, $maxDepth)
                 }
             } else if ($isObj) {
                 $return = new \stdclass();
-                if ($var instanceof \DateTime) {
-                    $return->__CLASS__ = "DateTime";
+                if ($var instanceof \DateTime || $var instanceof \DateTimeInterface) {
```
Ocramius: Can actually simply check against `DateTimeInterface`, since we don't support older PHP versions anymore

lib/Doctrine/Common/Util/Debug.php
```diff
@@ -126,12 +126,30 @@ public static function export($var, $maxDepth)
...
-                    foreach ($reflClass->getProperties() as $reflProperty) {
-                        $name  = $reflProperty->getName();
+                    $parsedAttributes = array();
+                    do {
+                        $currentClassName = $reflClass->getName();
 
-                        $reflProperty->setAccessible(true);
-                        $return->$name = self::export($reflProperty->getValue($var), $maxDepth - 1);
-                    }
+                        foreach ($reflClass->getProperties() as $reflProperty) {
```
Ocramius: Can you move this block to a private method?

tests/Doctrine/Tests/Common/Util/DebugTest.php
```diff
@@ -19,10 +19,23 @@ public function testExportObject( )
 
     public function testExportDateTime()
     {
-        $obj = new \DateTime( "2010-10-10 10:10:10" );
+        $obj = new \DateTime('2010-10-10 10:10:10', new \DateTimeZone('UTC'));
 
-        $var = Debug::export( $obj, 2 );
-        $this->assertEquals( "DateTime", $var->__CLASS__ );
+        $var = Debug::export($obj, 2);
+        $this->assertEquals('DateTime', $var->__CLASS__ );
```
Ocramius: Weird spaces at the end of these lines (leftovers from previous code)

tests/Doctrine/Tests/Common/Util/DebugTest.php
```diff
@@ -19,10 +19,23 @@ public function testExportObject( )
...
+        $this->assertEquals('2010-10-10T10:10:10+00:00', $var->date );
+    }
+
+    /**
+     * @requires PHP 5.5
+     */
+    public function testExportDateTimeImmutable()
+    {
+        $obj = new \DateTimeImmutable('2010-10-10 10:10:10', new \DateTimeZone('UTC'));
```
Ocramius: What happens with a different timezone?

Slamdunk:
Changes done.

In the last commit I applied "Replace Nested Conditional with Guard Clauses" pattern: if you don't want such a big change I can revert it (but if you ask me that I'll unfollow you in Twitter)

Slamdunk:
Note: I don't know if I did what you wanted saying "Can you move this block to a private method?", to me it was not strictly necessary, btw done it too.

Ocramius:
>  it was not strictly necessary, btw done it too.

The complexity of the method was starting to become a bit too much: splitting it simplifies things :)

lib/Doctrine/Common/Util/Debug.php
```diff
@@ -100,47 +100,85 @@ public static function export($var, $maxDepth)
...
+     * @return mixed
+     */
+    private static function fillReturnWithClassAttributes($var, \stdClass $return, $maxDepth)
+    {
+        $reflClass = ClassUtils::newReflectionObject($var);
+        $parsedAttributes = array();
+        do {
+            $currentClassName = $reflClass->getName();
+
+            foreach ($reflClass->getProperties() as $reflProperty) {
```
Ocramius: Last test scenario needed (although it's just to prevent regressions):

```php
class A
{
    private $a;
    protected $b;
    public $c;
}

class B extends A
{
    private $a;
    protected $b;
    public $c;
}
```

The resulting dump should show following properties, if I'm not mistaken:

 * `A#$a`
 * `B#$a`
 * `B#$b`
 * `B#$c`

Slamdunk:
Indeed there was a bug, fixed :P

Ocramius:
@Slamdunk awesome, thanks!

Issue #509, opened 2013-06-25
DCOM-204: AnnotationDriver cannot find classes inside Phar files


doctrinebot:
Jira issue originally created by user jgroth:

If an application is packaged inside a Phar file, the annotation driver cannot find any classes. The problem is in the method `AnnotationDriver::getAllClassNames()`

```
foreach ($iterator as $file) {
    $sourceFile = realpath($file[0]);

    require_once $sourceFile;

    $includedFiles[] = $sourceFile;
}
```

`$iterator` will hold all paths to the files found, however `realpath()` will return `false` since the files are inside a phar file. 

I have solved this locally in my application with the following, however I'm not sure if this is the right approach to solve the problem.

```
$sourceFile = str_replace('\\', '/', $file[0]);
if (!preg_match('#^phar://#i', $sourceFile)) {
    $sourceFile = realpath($sourceFile);
}
```

PR #732, opened 2016-08-05
Fixed path detection when running in Phar context

schroedan:
Because the issue https://github.com/doctrine/common/issues/509 wasn't really fixed by just omitting to call `realpath()` in Phar context, this will hopefully finally fix everything ;-)

The problem with the previous fix was that it didn't work with '/../' references inside. When they are compared against all declared classes (which have no '/../' in the file name) within the loop at the bottom of the method none of them will be found.

This fix will ensure that the file names can be compared in every context.

### DIFF
--- a/lib/Doctrine/Common/Persistence/Mapping/Driver/AnnotationDriver.php
+++ b/lib/Doctrine/Common/Persistence/Mapping/Driver/AnnotationDriver.php
@@ -270,6 +270,11 @@ abstract class AnnotationDriver implements MappingDriver
 
         $stack = [];
         $path = str_replace(['\\', '/'], DIRECTORY_SEPARATOR, substr($path, 7));
+
+        if (DIRECTORY_SEPARATOR === substr($path, 0, 1)) {
+            array_push($stack, '');
+        }
+
         foreach (explode(DIRECTORY_SEPARATOR, $path) as $folder) {
             if (0 === strlen($folder) || '.' === $folder) {
                 continue;


lib/Doctrine/Common/Persistence/Mapping/Driver/AnnotationDriver.php
```diff
@@ -253,4 +249,43 @@ public function getAllClassNames()
...
+     * Returns the canonicalized absolute pathname. This method is a Phar-safe version
+     * of the builtin `realpath()` function.
+     *
+     * @param string $path
+     *
+     * @return boolean|string
+     */
+    protected function realpath($path)
+    {
+        if ('' === \Phar::running()) {
```
Ocramius: This means that `ext-phar` is now a dependency of `doctrine/annotations`

Ocramius:
@pCoLaSD needs testing

pCoLaSD:
I removed the necessity of `ext-phar` by the check if the path starts with `phar://`. At the moment i have no clue how to write a unit test for that because I would have to simulate the Phar context.

lib/Doctrine/Common/Persistence/Mapping/Driver/AnnotationDriver.php
```diff
@@ -253,4 +249,43 @@ public function getAllClassNames()
...
+
+    /**
+     * Returns the canonicalized absolute pathname. This method is a Phar-safe version
+     * of the builtin `realpath()` function.
+     *
+     * @param string $path
+     *
+     * @return boolean|string
+     */
+    protected function realpath($path)
```
stof: should be private to avoid having to cover it with BC.

lib/Doctrine/Common/Persistence/Mapping/Driver/AnnotationDriver.php
```diff
@@ -253,4 +249,43 @@ public function getAllClassNames()
...
+    {
+        if ('phar://' !== substr($path, 0, 7)) {
```
stof: the phar stream wrapper is not the only one not supporting realpath. Otherwise stream wrappers might suffer the same.

What you should do is call realpath, and do the fallback logic when it returns ``false`` (and make the fallback logic support other cases than the phar stream wrapper)

stof:
Btw, making the logic handle any stream wrapper rather than just phar may make it easier to test it

Majkl578:
Hello,
Thanks.

PR #734, opened 2016-08-16
ProxyGenerator: Added support for PHP 7.1 features

Majkl578:
This was supposed to work without any change in the generator, but it's unfortunately not the case. There are multiple problems:
1) generating nullable type with FQCN is broken - generated code results in parse error;
2) reflection with nullable FQCN is broken - unable to generate proxy;
3) it's impossible to distinguish _nullable type with null default value_ and _non-nullable type with null default value_ - may change function signature in the proxy.

Forwarded upstream for more discussion to the PR where the changes were originally discussed and where you can also find more discussion with some other concerns: https://github.com/php/php-src/pull/2068#issuecomment-239983716
In case of no activity there I'll forward it to the bugs tracker soon or eventually php.internals.

(Travis still has old PHP 7.1 version without the changes, they'll appear in beta3.)

### DIFF
--- a/lib/Doctrine/Common/Proxy/ProxyGenerator.php
+++ b/lib/Doctrine/Common/Proxy/ProxyGenerator.php
@@ -787,7 +787,9 @@ EOT;
                 $cast       = in_array($fieldType, ['integer', 'smallint']) ? '(int) ' : '';
 
                 $methods .= '        if ($this->__isInitialized__ === false) {' . "\n";
-                $methods .= '            return ' . $cast . ' parent::' . $method->getName() . "();\n";
+                $methods .= '            ';
+                $methods .= $this->shouldProxiedMethodReturn($method) ? 'return ' : '';
+                $methods .= $cast . ' parent::' . $method->getName() . "();\n";
                 $methods .= '        }' . "\n\n";
             }
 
@@ -797,7 +799,9 @@ EOT;
             $methods .= "\n        \$this->__initializer__ "
                 . "&& \$this->__initializer__->__invoke(\$this, " . var_export($name, true)
                 . ", [" . $invokeParamsString . "]);"
-                . "\n\n        return parent::" . $name . '(' . $callParamsString . ');'
+                . "\n\n        "
+                . ($this->shouldProxiedMethodReturn($method) ? 'return ' : '')
+                . "parent::" . $name . '(' . $callParamsString . ');'
                 . "\n" . '    }' . "\n";
         }
 
@@ -935,18 +939,21 @@ EOT;
      */
     private function getParameterType(ClassMetadata $class, \ReflectionMethod $method, \ReflectionParameter $parameter)
     {
+        if (method_exists($parameter, 'hasType')) {
+            if ( ! $parameter->hasType()) {
+                return '';
+            }
 
-        // We need to pick the type hint class too
+            return $this->formatType($parameter->getType(), $parameter->getDeclaringFunction(), $parameter);
+        }
+
+        // For PHP 5.x, we need to pick the type hint in the old way (to be removed for PHP 7.0+)
         if ($parameter->isArray()) {
             return 'array';
         }
 
         if ($parameter->isCallable()) {
             return 'callable';
-        }
-
-        if (method_exists($parameter, 'hasType') && $parameter->hasType() && $parameter->getType()->isBuiltin()) {
-            return (string) $parameter->getType();
         }
 
         try {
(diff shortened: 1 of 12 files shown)


Majkl578:
Updated for PHP 7.1, but the Reflection API is still a mess IMHO. 😟
cc @Ocramius @trowski

Majkl578:
Added cd7fbb4 to show the inconsistent signature change for parameter without nullable type, but with null as default value. How should we handle this? I consider it a bug since the signature is different, but there is no easy way to support it properly (allowsNull returns true when default value is NULL).

Ocramius:
Note: I stopped working on this unit 7.1 is in RC phase, since the
ReflectionType API is basically useless in its current state. Posted about
it on thr mailing list multiple times, btw.

Ocramius:
Finished working on https://github.com/zendframework/zend-code/pull/87 - should be stable enough to be usable :-)

Ocramius:
Need checking of:

 * `public function foo(int $bar = null)` => `public function foo(?int $bar = null)`
 * `public function foo(iterable $bar)`
 * `public function foo($bar) : iterable`
 * `public function foo($bar) : void`

Majkl578:
Although 7.1 is RC2 already, Reflection is still a mess. Let's just wait for the resolution of https://github.com/php/php-src/pull/2136 and see what happens. It may probably also fix the nonsense with broken allowsNull. 🙏

Majkl578:
PR updated, now also includes support void return type and iterable pseudotype. Tested against PHP 7.1 RC3.

Ocramius:
@Majkl578 amazing! Reviewing now.

lib/Doctrine/Common/Proxy/ProxyGenerator.php
```diff
@@ -935,25 +939,49 @@ private function buildParametersString(ClassMetadata $class, \ReflectionMethod $
...
         }
 
         if ($parameter->isCallable()) {
-            return 'callable';
+            return $typeDeclaration . 'callable';
         }
 
-        if (method_exists($parameter, 'hasType') && $parameter->hasType() && $parameter->getType()->isBuiltin()) {
-            return (string) $parameter->getType();
+        if (method_exists($parameter, 'hasType')) {
```
Ocramius: This block is to be moved to a private method

lib/Doctrine/Common/Proxy/ProxyGenerator.php
```diff
@@ -935,25 +939,49 @@ private function buildParametersString(ClassMetadata $class, \ReflectionMethod $
      */
     private function getParameterType(ClassMetadata $class, \ReflectionMethod $method, \ReflectionParameter $parameter)
     {
+        $typeDeclaration = '';
```
Ocramius: This assignment and subsequent block is to be moved to a private method

lib/Doctrine/Common/Proxy/ProxyGenerator.php
```diff
@@ -1018,20 +1046,43 @@ private function getMethodReturnType(\ReflectionMethod $method)
 
         $returnType = $method->getReturnType();
 
+        $returnDeclaration = ': ';
```
Ocramius: To be moved to a private method (probably same method as block above)

tests/Doctrine/Tests/Common/Proxy/ProxyGeneratorTest.php
```diff
@@ -229,6 +229,92 @@ public function testClassWithReturnTypesOnProxiedMethods()
...
+        if (!class_exists('Doctrine\Tests\Common\ProxyProxy\__CG__\VoidReturnTypeClass', false)) {
+            $metadata = $this->createClassMetadata($className, ['id']);
+
+            $proxyGenerator = new ProxyGenerator(__DIR__ . '/generated', __NAMESPACE__ . 'Proxy', true);
+            $this->generateAndRequire($proxyGenerator, $metadata);
+        }
+
+        $classCode = file_get_contents(__DIR__ . '/generated/__CG__DoctrineTestsCommonProxyVoidReturnTypeClass.php');
+
+        $this->assertEquals(1, substr_count($classCode, 'function returnsVoid(): void'));
```
Ocramius: We should also verify that calling the method produces no return value. Without actually evaluating the code, we can't be sure that we support `void` correctly. Not sure if we should put it here or in https://github.com/doctrine/common/blob/670d814b6c8b5e8f39a2764d090463d39b3502c9/tests/Doctrine/Tests/Common/Proxy/ProxyLogicTest.php

Majkl578: What exactly do you mean? Since we can't put any return into the method that is to be proxied, just testing that such method returns NULL, i.e. `assertNull($proxy->returnsVoid())`?

tests/Doctrine/Tests/Common/Proxy/NullableTypeHintsClass.php
```diff
@@ -0,0 +1,45 @@
...
+
+    public function nullableTypeHintSelf(?self $param)
+    {
+    }
+
+    public function nullableTypeHintWithDefault(?int $param = 123)
+    {
+    }
+
+    public function nullableTypeHintWithDefaultNull(?int $param = null)
```
Majkl578: There is still no way to detect this correctly at runtime (makes one test fail).

tests/Doctrine/Tests/Common/Proxy/ProxyGeneratorTest.php
```diff
@@ -229,6 +231,94 @@ public function testClassWithReturnTypesOnProxiedMethods()
...
+            $this->generateAndRequire($proxyGenerator, $metadata);
+        }
+
+        $classCode = file_get_contents(__DIR__ . '/generated/__CG__DoctrineTestsCommonProxyNullableTypeHintsClass.php');
+
+        $this->assertEquals(1, substr_count($classCode, 'function nullableTypeHintInt(?int $param)'));
+        $this->assertEquals(1, substr_count($classCode, 'function nullableTypeHintObject(?\stdClass $param)'));
+        $this->assertEquals(1, substr_count($classCode, 'function nullableTypeHintSelf(?\\' . $className . ' $param)'));
+        $this->assertEquals(1, substr_count($classCode, 'function nullableTypeHintWithDefault(?int $param = 123)'));
+        $this->assertEquals(1, substr_count($classCode, 'function nullableTypeHintWithDefaultNull(?int $param = NULL)'));
```
Majkl578: This fails because Reflection API still doesn't provide proper way to detect this case (see method declaration above) and I don't see any straightforward workaround. Any advice how to handle this?

trowski: `ReflectionParameter::isOptional()` and `ReflectionParameter::getDefaultValue()`

Majkl578: @trowski Thanks, unfortunately that's not enough either, I used isDefaultValueAvailable which should be the same. See https://github.com/doctrine/common/pull/734/files#diff-ce03ab9b396edcbb313c54234c20e0deR1081

The following:
```php
public function foo(int $param = null)
```
would still generate:
```php
public function foo(?int $param = null)
```
which is not correct (at least in <7.1). And since we don't want any version checks, it's tricky...

tests/Doctrine/Tests/Common/Proxy/ProxyGeneratorTest.php
Ocramius: Something like `$proxy->increment(10);` and then `self::assertSame(10, $proxy->counter)`. Simply verifies that we're passing parameters.

tests/Doctrine/Tests/Common/Proxy/ProxyGeneratorTest.php
trowski: Is it needed for 7.1 to generate 7.0 compatible code? If so, you could special case nullable parameters with a null default value to always omit the preceding `?`. It may not be exactly as written, but it will be compatible.

Majkl578: @trowski That's a good question, probably not, but then someone could use 7.1 on CI server to warm-up cache (and generate proxies) and then 7.0 in production...

I already added special case for null default value, but then it has opposite effect.
I actually thought that this code would throw an error due to incompatible signatures:
```php
class Foo {
	function x(?int $x = NULL) {}
}
class Bar extends Foo {
	function x(int $x = NULL) {}
}
```
But surprisingly it doesn't, even though subclass removes nullability (question mark) from the type, something I'd not really expect. This is super-confusing, really something not well-thought-out in my opinion... 😳

trowski: `Bar::x(int $x = null)` is not is not removing nullability from `$x`, but rather just using older syntax to declare nullability. `?int $x = null` is semantically identical to `int $x = null`, so the definitions are compatible.

Majkl578:
- When default value is null, question mark is not added to ensure PHP 7.0 compatibility (see inline discussion above).
- Added test for void and calling parent with/without parameters.
- Hopefully fixed most of the CS issues (no code sniffer here to rescue me).

tests/Doctrine/Tests/Common/Proxy/ScalarTypeHintsClass.php
```diff
@@ -22,4 +22,12 @@ public function combinationOfTypeHintsAndNormal(\stdClass $a, $b, int $c)
     public function typeHintsWithVariadic(int ...$foo)
     {
     }
+
+    public function withDefaultValue(int $foo = 123)
```
Ocramius: Oooh, shiny! I'll have to test this on my libs too.

Ocramius:
@Majkl578 excellent work! The only blocker for getting this merged is having `7.1.0-RC3` on travis, since there were some changes in the API there.

Ocramius:
Actually, never mind, tested locally, and everything works! Merging, thanks!

PR #735, opened 2016-08-23
Add hasEmbeddedClass method to ClassMetadata interface

foaly-nr1:
Just like `hasField()` checks `$fieldMappings` and `hasAssociation()` checks `$associationMappings`, the proposed `hasEmbeddedClass()` will check `$embeddedClasses` for the field in question.

Basically the missing variable in this equation:

``` php
$this->assertFieldNotMapped($fieldName) === !($this->hasField($fieldName) && $this->fieldName($fieldName) && $this->hasEmbeddedClass($fieldName))
```

### DIFF
--- a/lib/Doctrine/Common/Persistence/Mapping/ClassMetadata.php
+++ b/lib/Doctrine/Common/Persistence/Mapping/ClassMetadata.php
@@ -98,6 +98,15 @@ interface ClassMetadata
     public function isCollectionValuedAssociation($fieldName);
 
     /**
+     * Checks if the given field has an embedded class for this class.
+     *
+     * @param string $fieldName
+     *
+     * @return boolean
+     */
+    public function hasEmbeddedClass($fieldName);
+
+    /**
      * A numerically indexed list of field names of this persistent class.
      *
      * This array includes identifier fields if present on this class.


Majkl578:
This would be a BC break - it could only be acceptable for 3.0 release. You should also update tests to work with this new method.

lcobucci:
Embeddables will being reviewed/rewritten on ORM 3. @guilhermeblanco is handling this so maybe he can give more context to you.

malarzm:
I'd just add a note that `hasEmbeddedClass` seem to be ORM specific, for instance MongoDB ODM treats both `@EmbedOne` and `@EmbedMany` as associations

guilhermeblanco:
This method (and others) should disappear in the future.
Ideally what you want to check is if the property is defined or not... checking the property type is something that doesn't make any sense in the internal metadata code, but the consumer.

```php
$class->getProperty($fieldName) instanceof EmbeddableMetadata
```

The same should be valid for `FieldMetadata`, `AssociationMetadata` (and its subclasses, such as `ToOneAssociationMetadata`, `OneToManyAssociationMetadata`, etc).

Also, checking property existence should be:

```php
$class->hasProperty($fieldName)
```

I don't see the need to do anything else besides that. So -1 on this PR.

Ocramius:
Closing as per @guilhermeblanco's https://github.com/doctrine/common/pull/735#issuecomment-295067163

PR #737, opened 2016-08-24
Ensure a leading backslash (and only one) is returned

### LINKED ISSUE #736: Proxies / PHP7 return types: unexpected '\' (T_NS_SEPARATOR)
Hi there,

It looks like the ProxyGenerator doesn't correctly generate the return type on a namespaced class. Here's my example:

``` php
# src/My/Vendor/Entity/MyEntity.php
namespace My\Vendor\Entity;
class MyEntity {
// ...
    public function getOtherEntity() : OtherEntity {
        return $this->otherEntity;
    }
}
```

And the generated proxy code:

``` php
# var/cache/dev/doctrine/orm/Proxies/__CG__MyVendorEntityMyEntity.php
class MyEntity extends My\Vendor\Entity\MyEntity implements \Doctrine\ORM\Proxy\Proxy {
// ...
    public function getOtherEntity() : \\My\Vendor\Entity\OtherEntity {
        $this->__initializer__ && $this->__initializer__->__invoke($this, 'getOtherEntity', []);

        return parent::getOtherEntity();

    }
}
```

Which results in the following syntax error:

```
Parse error: syntax error, unexpected '\' (T_NS_SEPARATOR), expecting identifier (T_STRING)
```

Indeed a double leading backslash is generated on top of the return type.

Ping #376

bpolaszek:
... in the return type template.
Fixes #736

### DIFF
--- a/lib/Doctrine/Common/Proxy/ProxyGenerator.php
+++ b/lib/Doctrine/Common/Proxy/ProxyGenerator.php
@@ -1032,6 +1032,6 @@ EOT;
             return ': \\' . $method->getDeclaringClass()->getParentClass()->getName();
         }
 
-        return ': \\' . (string) $returnType;
+        return ': \\' . ltrim((string) $returnType, '\\');
     }
 }
--- a/tests/Doctrine/Tests/Common/Proxy/ProxyGeneratorTest.php
+++ b/tests/Doctrine/Tests/Common/Proxy/ProxyGeneratorTest.php
@@ -222,6 +222,7 @@ class ProxyGeneratorTest extends PHPUnit_Framework_TestCase
         $classCode = file_get_contents(__DIR__ . '/generated/__CG__DoctrineTestsCommonProxyReturnTypesClass.php');
 
         $this->assertEquals(1, substr_count($classCode, 'function returnsClass(): \stdClass'));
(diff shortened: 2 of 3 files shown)


Ocramius:
@bpolaszek can you please add a test case that shows this behavior? An additional test asset should be sufficient.

Did you by chance/accident run the tests on PHP 7.1.0-beta3? That beta is broken...

bpolaszek:
@Ocramius Good catch, I was on PHP 7.1.0-beta3 indeed :)
I added an additionnal asset. The tests succeed on PHP 7.0.4 as well.

lib/Doctrine/Common/Proxy/ProxyGenerator.php
```diff
@@ -1032,6 +1032,6 @@ private function getMethodReturnType(\ReflectionMethod $method)
             return ': \\' . $method->getDeclaringClass()->getParentClass()->getName();
         }
 
-        return ': \\' . (string) $returnType;
+        return ': \\' . ltrim((string) $returnType, '\\');
```
Ocramius: The tests pass also without this change.

Ocramius:
Closing then: 7.1.0-beta3 is bugged as per http://marc.info/?t=147145076000010&r=1&w=2, https://github.com/php/php-src/pull/2068 and other reported bugs.

We'll see how 7.1.0-RC1 goes.

PR #741, opened 2016-09-22
Adding a lost parameter in ObjectRepository

andrewprofile:
Based on the documentation EntityRepository I noticed there findOneBy method has two parameters. But when I tried to use it with the second parameter, my IDE PhpStorm reported the mistake that this method has only one parameter. This is due to the fact that it implements interface ObjectRepository without this parameter. Therefore, I decided to add this parameter for better code completion.

Ps. Sory for my bad english

### DIFF
--- a/lib/Doctrine/Common/Persistence/ObjectRepository.php
+++ b/lib/Doctrine/Common/Persistence/ObjectRepository.php
@@ -66,11 +66,12 @@ interface ObjectRepository
     /**
      * Finds a single object by a set of criteria.
      *
-     * @param array $criteria The criteria.
+     * @param array      $criteria The criteria.
+     * @param array|null $orderBy
      *
      * @return object|null The object.
      */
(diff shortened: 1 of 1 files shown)


Ocramius:
This can't be implemented, as it is a BC break in the interface, sorry...

andrewprofile:
Yes, You are right it will BC break for Doctrine 2.1.x - 2.2. I'm sorry, I did not check it before.

stof:
No, it breaks BC for anyone implementing the interface today, noit only old ORM versions (the ORM is *not* the only implementer)

andrewprofile:
Ok, I understand, thanks. Sorry for the mistake.

Ocramius:
@andrewprofile we have our hands tied on basically anything that involves changing interfaces, sorry :-\

andrewprofile:
Ok, Thanks for your feedback.

PR #743, opened 2016-10-06
Fix parameter/return type validation for interfaces (introduced in #734)

Majkl578:
Something that unfortunately got through in #734 after recent refactoring and caused UnexpectedValueException when inteface was used in parameter or return type. (Original piece of code used to catch ReflectionException instead.)
Noticed when testing against code-base with `getFoo() : Collection`.


Ocramius:
@Majkl578 thanks for keeping an eye on it! Well spotted!

Issue #748, opened 2016-11-22
AnnotationDriver is not loading entities from symlinked directories


aurelijusrozenas:
Entity directory structure (AppBundle/Entity)
 - Blog (directory)
    - Post.php
 - Blog2 (symlink)
    - Post2.php

Mapping loaded from AnnotationDriver (e.g. via symfony's `bin/console doctrine:mapping:info`) will display info for Post but not for Post2 because `Blog2` is symlinked.

`\RecursiveDirectoryIterator::FOLLOW_SYMLINKS` flag need to be passed to `RecursiveDirectoryIterator` to fix this.

To reproduce this in symfony:
```bash
symfony new symlinkedEntities
composer install
bin/console doctrine:gen:entity --entity=AppBundle:Blog/Post --format=annotation --fields="title:string(length=255 nullable=true unique=true)" --no-interaction
# set DB params adn then
bin/console doc:data:create
# works
bin/console doctrine:gen:entities AppBundle:Blog

# does not work
mv src/AppBundle/Entity/Blog .
ln -s FULL_PATH/symlinkedEntities/Blog src/AppBundle/Entity/Blog
bin/console doctrine:gen:entities AppBundle:Blog
```

>Generating entities for namespace "AppBundle\Entity\Blog"
  [RuntimeException]
  Namespace "AppBundle\Entity\Blog" does not contain any mapped entities.

PR #749, opened 2016-11-22
Fixed AnnotationDriver to include entity files from symlinked directories.

aurelijusrozenas:
Fixes https://github.com/doctrine/common/issues/748

Fix is easy one-liner. However to make this code great, I need some guidance.

## What branch to merge to?
Branched from master and made PR back to master.

## Other drivers?
This is only tested for AnnotationDriver, other drivers should not be affected.

### DIFF
--- a/tests/Doctrine/Tests/Common/Persistence/Mapping/AnnotationDriverTest.php
+++ b/tests/Doctrine/Tests/Common/Persistence/Mapping/AnnotationDriverTest.php
@@ -21,7 +21,11 @@ class AnnotationDriverTest extends \PHPUnit_Framework_TestCase
 
         $classes = $driver->getAllClassNames();
 
-        $this->assertEquals([SubDirTestClass::class, SymlinkedTestClass::class, TestClass::class], $classes);
+        $expectedClasses = [SubDirTestClass::class, SymlinkedTestClass::class, TestClass::class];
+        sort($expectedClasses);
+        sort($classes);
+
+        $this->assertEquals($expectedClasses, $classes);
     }
 
     private function makeSymlink()
--- a/tests/Doctrine/Tests/Common/Persistence/Mapping/AnnotationDriverTest.php
+++ b/tests/Doctrine/Tests/Common/Persistence/Mapping/AnnotationDriverTest.php
@@ -26,7 +26,7 @@ class AnnotationDriverTest extends \PHPUnit_Framework_TestCase
 
     private function makeSymlink()
     {
-        $symlinkTarget = __DIR__ . '/../SubDirForSymlink';
+        $symlinkTarget = __DIR__ . '/_files/SubDirForSymlink';
         $symlinkPath = __DIR__ . '/_files/annotation/SubDirForSymlink';
         if (!is_link($symlinkPath)) {
             if (file_exists($symlinkPath)) {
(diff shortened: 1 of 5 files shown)


stof:
if other drivers are also be impacted, they should also be fixed. Supporting symlinks in no driver could be considered OK (it is a maintainer choice then). But being inconsistent among drivers, with support in only some of them, would just be a WTF for users.

aurelijusrozenas:
Sure thing. Any ideas about the tests?

Ocramius:
Don't really want symlink support, especially if we consider shady platforms like Windows, where symlinks are Schrödinger's symlinks...

Introducing a documentation change+limitation, or the ability to pass the iterator directly to the driver would be better solutions, in my opinion.

aurelijusrozenas:
I'll look into it.

ShinDarth:
are there any news about this?

aurelijusrozenas:
If we are talking about symlinked entity directories, other drivers are not impacted because they load information from other directories. Right?
Loading metadata information from symlinked config dirs could be separate feature, I did not do anything for that.
@Ocramius not sure about windows. I it would be interesting to see if my updated test works on them.

tests/Doctrine/Tests/Common/Persistence/Mapping/AnnotationDriverTest.php
```diff
@@ -6,18 +6,39 @@
...
+        sort($classes);
+
+        $this->assertEquals($expectedClasses, $classes);
+    }
+
+    private function makeSymlink()
+    {
+        $symlinkTarget = __DIR__ . '/_files/SubDirForSymlink';
+        $symlinkPath = __DIR__ . '/_files/annotation/SubDirForSymlink';
+        if (!is_link($symlinkPath)) {
```
phansys: Please, respect CS (` ! is_link()`).

tests/Doctrine/Tests/Common/Persistence/Mapping/AnnotationDriverTest.php
```diff
@@ -6,18 +6,39 @@
...
+            if (file_exists($symlinkPath)) {
+                throw new \Exception(sprintf('Remove file/dir "%s".', $symlinkPath));
+            } else {
```
phansys: This `else` condition could be omitted given the previous uncaught exception.

tests/Doctrine/Tests/Common/Persistence/Mapping/AnnotationDriverTest.php
```diff
@@ -6,18 +6,38 @@
...
+
+    private function makeSubDirSymlink()
+    {
+        $symlinkTarget = __DIR__ . '/_files/SubDirForSymlink';
+        $symlinkPath = __DIR__ . '/_files/annotation/SubDirForSymlink';
+        if (! is_link($symlinkPath)) {
+            if (file_exists($symlinkPath)) {
+                throw new \Exception(sprintf('Remove file/dir "%s".', $symlinkPath));
+            }
+            symlink($symlinkTarget, $symlinkPath);
```
Ocramius: How does this test behave on a windows system?

aurelijusrozenas: Creating symlink on windows requires "Run as administrator" permissions, then tests run fine. Otherwise it gives "1) Doctrine\Tests\Common\Persistence\Mapping\Annota
symlink(): Cannot create symlink, error code(1314)".

Majkl578:
Hello,
Thanks.

PR #756, opened 2017-01-02
Removes support for PHP 5.x

railto:
Removes support for PHP 5.x from composer.json and .travis.yml

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,15 +10,13 @@ php:
   - 7.0
   - 7.1
 
-  matrix:
+matrix:
   fast_finish: true
   include:
     - php: hhvm
       sudo: true
       dist: trusty
       group: edge # until the next update
-  allow_failures:
-    - php: hhvm
 
 before_script:
   - composer --prefer-source install


stof:
And this work should start on the ORM and ODM side (otherwise they would not be able to require using new features)

Ocramius:
@stof ODM doesn't depend on any new stuff AFAIK /cc @alcaeus

.travis.yml
```diff
@@ -7,11 +7,14 @@ cache:
...
 
 php:
-  - 5.6
   - 7.0
   - 7.1
   - hhvm
 
+matrix:
+    allow_failures:
+        - php: hhvm
```
Ocramius: This is not a solution: check if a newer HHVM version is available - HHVM now support PHP7 stuff.

railto: Apologies, was bringing it in line with the other repos, will read up on how to force a version of HHVM as don't use it much myself at present.

.travis.yml
```diff
@@ -10,15 +10,13 @@ php:
   - 7.0
   - 7.1
 
-  matrix:
+matrix:
   fast_finish: true
   include:
```
railto: This brings in the current version of HHVM, 3.17, however is still throwing the same error with 
```This package requires php ^7.0 but your HHVM version does not satisfy that requirement.```
This error is also happening on other repos like the primary doctrine2 repo per this test - https://travis-ci.org/doctrine/doctrine2/jobs/188191005

Not sure what else I can do with this as HHVM 3.17 is not reporting to be PHP 7.x compatible per composer.

alcaeus:
@stof @Ocramius ODM 2.0 (which will use the new `ext-mongo` extension) will be PHP 7.1 only - I might even go as far as kicking PHP 5.x support for ODM 1.2 (which we might do before tagging 2.0). Either way, I have no problem with doctrine/common dropping 5.x support for 2.8.

.travis.yml
Ocramius: @fredemmott can you eventually give us some insight on this?

fredemmott: You need hhvm.php7.all=true in the ini file; I don't know how to do this on Travis. Composer might also need some poking to understand what that does.

cc @paulbiss, @aorenste for that - also, it might be time to flip the default of that ini setting.

railto:
Closing as I do not have the ability to get this to work with HHVM tests

Ocramius:
Reopening: still needs to happen, even if you don't work on the PR yourself.

Majkl578:
I tried this using `--ignore-platform-reqs` composer switch, but it triggered bunch of test failures:
https://travis-ci.org/Majkl578/doctrine-common/jobs/198386782
Might be because my build used 3.17.2 whereas doctrine/common uses 3.6 (although both specify just `hhvm` dependency).

.travis.yml
lcobucci: @railto on doctrine/doctrine2#6288 I've managed to get the test suite running (but HHVM fails badly in terms of PHP compatibility that I gave up 😭).

What I've learned:

-  `dist` should be defined on the root level
- we must set `hhvm.php7.all=1` AND `hhvm.php7.scalar_types=0` otherwise it doesn't work properly (HHVM simply enables strict mode for all files)

Majkl578:
Handled in #785.

malarzm:
Closing in favour of #785

Issue #761, opened 2017-01-13
ProxyGenerator::isShortIdentifierGetter in PHP 7.1 does not support nullable types


aurimasniekis:
ProxyGenerator::isShortIdentifierGetter does not work with PHP7.1 nullable return types.

Does not work
```
public function getId() : ?int
```

Works

```
public function getId() : int
```

If I understood correctly the `ProxyGenerator::PATTERN_MATCH_ID_METHOD` should be updated to include question mark in return type and test cases to have case with nullable return type.

PR #762, opened 2017-01-13
#761 Added support for PHP 7.1 identifier nullable return type

aurimasniekis:
The only problem I see now. That doctrine/common does not yet support PHP 7.1 in `composer.json`. So I dont know how to release this because test will fail under <7.1 version

This is fix for #761

### DIFF
--- a/lib/Doctrine/Common/Proxy/ProxyGenerator.php
+++ b/lib/Doctrine/Common/Proxy/ProxyGenerator.php
@@ -37,7 +37,7 @@ class ProxyGenerator
      * Used to match very simple id methods that don't need
      * to be decorated since the identifier is known.
      */
-    const PATTERN_MATCH_ID_METHOD = '((public\s+)?(function\s+%s\s*\(\)\s*)\s*(?::\s*\\\\?[a-z_\x7f-\xff][\w\x7f-\xff]*(?:\\\\[a-z_\x7f-\xff][\w\x7f-\xff]*)*\s*)?{\s*return\s*\$this->%s;\s*})i';
+    const PATTERN_MATCH_ID_METHOD = '((public\s+)?(function\s+%s\s*\(\)\s*)\s*(?::\s*\?*\\\\?[a-z_\x7f-\xff][\w\x7f-\xff]*(?:\\\\[a-z_\x7f-\xff][\w\x7f-\xff]*)*\s*)?{\s*return\s*\$this->%s;\s*})i';
 
     /**
      * The namespace that contains all proxy classes.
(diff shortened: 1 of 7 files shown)


lib/Doctrine/Common/Proxy/ProxyGenerator.php
```diff
@@ -37,7 +37,7 @@ class ProxyGenerator
      * Used to match very simple id methods that don't need
      * to be decorated since the identifier is known.
      */
-    const PATTERN_MATCH_ID_METHOD = '((public\s+)?(function\s+%s\s*\(\)\s*)\s*(?::\s*\\\\?[a-z_\x7f-\xff][\w\x7f-\xff]*(?:\\\\[a-z_\x7f-\xff][\w\x7f-\xff]*)*\s*)?{\s*return\s*\$this->%s;\s*})i';
+    const PATTERN_MATCH_ID_METHOD = '((public\s+)?(function\s+%s\s*\(\)\s*)\s*(?::\s*\?*\\\\?[a-z_\x7f-\xff][\w\x7f-\xff]*(?:\\\\[a-z_\x7f-\xff][\w\x7f-\xff]*)*\s*)?{\s*return\s*\$this->%s;\s*})i';
```
Ocramius: `\?{0,1}`, not `*`

tests/Doctrine/Tests/Common/Proxy/LazyLoadableObjectWithTypehints.php
```diff
@@ -90,5 +93,10 @@ public function getIdentifierFieldReturnClassOneLetter(): A
         return $this->identifierFieldReturnClassOneLetter;
     }
 
+    public function getIdentifierFieldReturnClassOneLetterNullable(): ?A
```
Ocramius: This one probably needs to be moved to a separate class, to avoid having the parser choke when importing the test assets on pre-7.1

lib/Doctrine/Common/Proxy/ProxyGenerator.php
stof: should not be ``\?*`` but ``\??\s*`` instead. We must allow a single ``?`` char, and spaces after it

tests/Doctrine/Tests/Common/Proxy/LazyLoadableObjectWithTypehints.php
stof: this must be done in a separate class, in a test being skipped on PHP < 7.1. Otherwise, it breaks the testsuite for PHP 7.0

aurimasniekis: Yes this, what I mentioned in description ok will do it 👍

Ocramius:
Moved to #763

PR #763, opened 2017-01-13
Fix #761, #762: support nullable identifier hint in proxies

### LINKED ISSUE #761: ProxyGenerator::isShortIdentifierGetter in PHP 7.1 does not support nullable types

Ocramius:
Fixes #761
Fixes #762
Replaces #762 (includes its commits)


tests/Doctrine/Tests/Common/Proxy/LazyLoadableObjectWithNullableTypehints.php
```diff
@@ -0,0 +1,50 @@
...
+
+    /** @var \stdClass */
+    private $identifierFieldReturnClassOneLetterNullableWithSpace;
+
+    public function getIdentifierFieldReturnClassOneLetterNullable(): ?A
+    {
+        return $this->identifierFieldReturnClassOneLetterNullable;
+    }
+
+    public function getIdentifierFieldReturnClassOneLetterNullableWithSpace(): ? A
```
Soullivaneuh: Don't know what is the correct syntax, but you set `? A` here and `?A`.

I suppose this is not wanted. :wink:

Ocramius: Space is intentional, since both approaches are valid/accepted

Soullivaneuh: Oh indeed.

But why do you have to test both syntaxes? The behavior is same, am I wrong?

Ocramius: Yes, but the code analyzer is regex-based (yeah, I know :S)

Soullivaneuh: > Yes, but the code analyzer is regex-based

Damn me, indeed!

> (yeah, I know :S)

Regex is good, eat it! :trollface:

Ocramius:
Backported into `2.7` in doctrine/common@930297026c8009a567ac051fd545bf6124150347

PR #769, opened 2017-02-01
Specify array content type for ClassMetadataFactory::getAllMetadata

Soullivaneuh:
When the array contains only instances of `ClassMetadata`, then we can specify it on PHPdoc return tag for better IDE completion and documentation.


stof:
:+1:

Ocramius:
👍

PR #771, opened 2017-02-03
ProxyGenerator: Disallow creating proxies for abstract and final classes

### LINKED ISSUE #770: ProxyFactory may generate invalid proxies with no error
Follow up of https://github.com/symfony/symfony/issues/21509

If for example the entity is final, as the generated proxy make use of inheritance, this throw a `FatalError` when the proxy class is being used. However, this error is thrown at runtime when the proxy is being used, not when it is generated. WDYT of adding a simple `include_once` on the generated proxy when creating it, to throw an error as early as possible?

Majkl578:
Fixes #770
Related to symfony/symfony#21509

_This could theoretically be a BC break, but I can't imagine any such scenario._

### DIFF
--- a/lib/Doctrine/Common/Proxy/ProxyGenerator.php
+++ b/lib/Doctrine/Common/Proxy/ProxyGenerator.php
@@ -256,10 +256,13 @@ class <proxyShortClassName> extends \<className> implements \<baseProxyInterface
      * @param \Doctrine\Common\Persistence\Mapping\ClassMetadata $class    Metadata for the original class.
      * @param string|bool                                        $fileName Filename (full path) for the generated class. If none is given, eval() is used.
      *
+     * @throws InvalidArgumentException
      * @throws UnexpectedValueException
      */
     public function generateProxyClass(ClassMetadata $class, $fileName = false)
     {
+        $this->verifyClassCanBeProxied($class);
+
         preg_match_all('(<([a-zA-Z]+)>)', $this->proxyClassTemplate, $placeholderMatches);
 
         $placeholderMatches = array_combine($placeholderMatches[0], $placeholderMatches[1]);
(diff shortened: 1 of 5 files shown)


lib/Doctrine/Common/Proxy/ProxyGenerator.php
```diff
@@ -260,6 +260,8 @@ public function setProxyClassTemplate($proxyClassTemplate)
      */
     public function generateProxyClass(ClassMetadata $class, $fileName = false)
     {
+        $this->verifyClassValidnessForProxy($class);
```
Ocramius: `verifyClassCanBeProxied`

lib/Doctrine/Common/Proxy/ProxyGenerator.php
```diff
@@ -306,6 +308,17 @@ public function generateProxyClass(ClassMetadata $class, $fileName = false)
         rename($tmpFileName, $fileName);
     }
 
+    private function verifyClassValidnessForProxy(ClassMetadata $class)
```
Ocramius: Please add relevant `@throws` annotations in the private/public API of the class being affected by the change

lib/Doctrine/Common/Proxy/ProxyGenerator.php
Majkl578: I admit I was out of fantasy. 😇 Renamed.

Ocramius:
@Majkl578 awesome, thanks!

theofidry:
👌

pamil:
Unfortunately, this breaks Sylius (already fixed in dev-master) - looks like we don't really use an entity which was final, so it has started to break after an update to 2.8. If anyone bumped onto this issue as well, just use this workaround https://github.com/Sylius/SyliusElasticSearchPlugin/pull/55/files#diff-b5d0ee8c97c7abd7e3fa29b9a27d1780R41 and it works just fine :)

bendavies:
Yes, this breaks using https://github.com/moneyphp/money/ as an embedded.

Ocramius:
Related: https://github.com/doctrine/doctrine2/issues/6625

PR #775, opened 2017-02-13
Use __call() to prevent calling unsupported methods on StaticReflectionClass

alexpott:
PHP7 recently merged and then reverted https://github.com/php/php-src/pull/2363. This caused problems for StaticReflectionClass because the signature of StaticReflectionClass::newInstance() is changed. The commit remains in PHP's HEAD so at some point this will break.

### DIFF
--- a/lib/Doctrine/Common/Reflection/StaticReflectionClass.php
+++ b/lib/Doctrine/Common/Reflection/StaticReflectionClass.php
@@ -88,338 +88,67 @@ class StaticReflectionClass extends ReflectionClass
     }
 
     /**
-     * {@inheritDoc}
+     * Prevents calling unsupported methods.
+     *
+     * __call() is used so that PHP can change the signature of these methods without breaking this class. For example,
+     * https://github.com/php/php-src/pull/2363.
+     *
+     * @param string $method
+     * @param array  $args
+     *
+     * @return mixed
+     *
+     * @throws \ReflectionException
      */
-    public static function export($argument, $return = false)
+    public function __call($name, $arguments)
     {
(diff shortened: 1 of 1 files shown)


lib/Doctrine/Common/Reflection/StaticReflectionClass.php
```diff
@@ -88,337 +88,19 @@ public function getProperty($name)
...
+     *
+     * @param string $method
+     * @param array  $args
+     *
+     * @return mixed
+     *
+     * @throws \ReflectionException
      */
-    public function setStaticPropertyValue($name, $value)
+    public function __call($name, $arguments)
```
Ocramius: This wouldn't be called for public API that is defined on the parent class.

alexpott: Of course :( brain fart - sorry. Thanks for the review.

alexpott: We could just remove the methods and let it break horribly?

alexpott: That does not look like a good solution since you'll just get a "Internal error: Failed to retrieve the reflection object" error - which is not very helpful.

dawehner:
Couldn't we also have a gigantic ```if``` in that file and define the same class twice? I know this is horrible :(

Ocramius:
The best solution, for now, is to simply blacklist any PHP release affected by the bug. Consider figuring that out first, and a simple `composer.json` addition would do the trick.

PR #778, opened 2017-03-03
Fix return type docblock for dispatchEvent

gadelat:


Ocramius:
@gadelat LGTM, thanks! :+1:

PR #780, opened 2017-03-09
Quick-fix to avoid call to deprecated function.

fridde:
LifecycleEventArgs::getEntity() is deprecated and should be replaced by getObject() to maintain consistency.


stof:
👍

Ocramius:
LGTM :+1:

Before I merge, is this already covered by tests?

stof:
@Ocramius none of the event objects are covered by tests (but this particular issue would not have been spotted by tests, as deprecated methods are only documented as such, and don't report their usage programmatically)

Ocramius:
:ship:

Ocramius:
Backported via doctrine/common@ef719c984ee6f4790aea80ee20712c2709c104bb

PR #785, opened 2017-04-15
Drop PHP 5.x & 7.0 support

Majkl578:
Supersedes and fixes #756.


.travis.yml
```diff
@@ -20,7 +19,8 @@ matrix:
       env: PHPSTAN=1
 
 before_script:
-  - composer --prefer-source install
+  - if [[ $TRAVIS_PHP_VERSION = hhvm ]]; then composer --ignore-platform-reqs --prefer-source install; fi
```
Ocramius: `--ignore-platform-reqs` is not acceptable here

Majkl578: Fine, I tried. 😊 Damn I hate HHVM for this versioning hell... When switched to php7 mode (`hhvm --php -d hhvm.php7.all=true`), it works, but it's messy...

Majkl578:
Okay, HHVM will just *not* work in PHP 7 mode at all... It crashes Composer entirely in all versions (Symfony Finder specifically).
Suggestions?

Majkl578:
See https://github.com/facebook/hhvm/issues/7198#issuecomment-284493814.

szymach:
Hey, I am curious whether support for HHVM is actually worth the effort? Perhaps add it as an allowed failure until the Facebook team resolves the issue?

Majkl578:
Maybe, but it's not up to me to decide.
HHVM support was always pain in the ass, i.e. Nette Framework dropped it because it was too broken/incompatible with PHP.

szymach:
Well, according Salded in the [issue](https://github.com/composer/composer/issues/6230) you have provided, there's less people using HHVM and Composer, than PHP 5.3 (~0.4%). If I may be so bold, I would suggest sidelining (if not dropping altogether) support for it, since it only slows down development progress without being actually used by any substantial number of people. 

I know, it is not your call, but maybe someone else in the Doctrine team will consider this.

Ocramius:
I think @szymach's observation makes it clear: supporting HHVM is not worth it, and won't be worth it any soon due to the amount of effort needed to do it, while a lot of its PHP5 compatibility issues remain unpatched (and aren't in the project's focus either, as it seems).

We can drop HHVM if we do an announcement, since we did indeed guarantee compatibility (see http://www.doctrine-project.org/2013/12/23/our-hhvm-roadmap.html )

Majkl578:
@Ocramius @lcobucci Before I rebase this, are we going to support 7.0 or 7.1?

Ocramius:
@Majkl578 7.1

Majkl578:
Roger, PR updated.

Ocramius:
:+1:

Issue #787, opened 2017-04-27
ProxyGenerator ignores public embedded fields


mkosiedowski:
Getting public embeddables on lazy-loaded entity which is not initialized does not trigger entity initialization.

Bug is in `\Doctrine\Common\Proxy\ProxyGenerator::getLazyLoadedPublicProperties` in code:
```php
if (($class->hasField($name) || $class->hasAssociation($name)) && ! $class->isIdentifier($name)) {
    $properties[$name] = $defaultProperties[$name];
}
```
This code only checks for fields and associations, but not for embeddables. For now the interface `\Doctrine\Common\Persistence\Mapping\ClassMetadata` doesn't even provide any function to check if a field is an embeddable object.

mkosiedowski:
As agreed in https://github.com/doctrine/common/pull/788 this should be fixed in ORM

Ocramius:
Handled in https://github.com/doctrine/doctrine2/pull/6462

PR #788, opened 2017-04-27
Initialize proxy when fetching public embeddable

mkosiedowski:
This fixes issue https://github.com/doctrine/common/issues/787
This change introduces new interface method which will have to be implemented in `\Doctrine\ORM\Mapping\ClassMetadataInfo`

### DIFF
--- a/lib/Doctrine/Common/Persistence/Mapping/ClassMetadata.php
+++ b/lib/Doctrine/Common/Persistence/Mapping/ClassMetadata.php
@@ -80,6 +80,15 @@ interface ClassMetadata
     public function hasAssociation($fieldName);
 
     /**
+     * Checks if the given field is an embedded class for this class.
+     *
+     * @param string $fieldName
+     *
+     * @return boolean
+     */
+    public function hasEmbeddedField($fieldName);
+
+    /**
      * Checks if the given field is a mapped single valued association for this class.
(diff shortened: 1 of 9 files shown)


malarzm:
@mkosiedowski unfortunately the new method can not be introduced as it is a BC break (also please see https://github.com/doctrine/common/pull/735). I think the best to solve the issue would be an additional check against ORM metadata implementation and using its method then to determine if embedded field is there

mkosiedowski:
@malarzm I see... I created a dirty hack for this. I hate dirty hacks so I will totally understand if you reject it anyway ;)

malarzm:
This is dirty hack I was thinking about, however for accepting or rejecting it it's better to ping @Ocramius ;)

mkosiedowski:
Hmm... I am now wondering if this bug actually shouldn't be solved in ORM implementation... Maybe `hasField` should return `true` also for embedded fields... This will require some refactoring in ORM code, but this might be better path to solve this issue.

lcobucci:
@mkosiedowski indeed, embeddable concept is not the same in ORM and ODM and so it makes more sense to solve this in the ORM.

Ocramius:
Handled in Handled in https://github.com/doctrine/doctrine2/pull/6462

PR #792, opened 2017-04-30
Fix __call exception when no metadata has been set

### LINKED ISSUE #139: generation of error message fails under certain circumstances
in `PersistentObject::__call` method, it is possible that `$command` is not `set`, `get` or `add`. If additonally `$this->cm` is null, you will get a `Call to a member function getName() on a non-object` error message. Instead, one should check if `$this->cm` is available before throwing.

ostrolucky:
This makes sure $cm is set before trying to call it in throw block
closes #139

### DIFF
--- a/lib/Doctrine/Common/Persistence/PersistentObject.php
+++ b/lib/Doctrine/Common/Persistence/PersistentObject.php
@@ -119,8 +119,6 @@ abstract class PersistentObject implements ObjectManagerAware
      */
     private function set($field, $args)
     {
-        $this->initializeDoctrine();
-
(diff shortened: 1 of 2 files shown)


ostrolucky:
What test changes? You mean writing new test?

Ocramius:
Yup. As much as I'm sorry for no tests existing so far, new changes require
to be validated by tests.

ostrolucky:
test added

Ocramius:
Thanks @ostrolucky!

Ocramius:
Backported via doctrine/common@407abe1ee13e1ed2436e6802987e51307f9ed86f

PR #802, opened 2017-06-24
Fix #386 - Debug export does not include stdClass properties

### LINKED ISSUE #386: Debug export does not include stdClass properties
Example to reproduce:

``` php
$obj = new \stdClass;
$obj->foo = "bar";
$obj->bar = 1234;
$var = \Doctrine\Common\Util\Debug::export($obj, 2);

class a { public $foo; public $bar;}
$obj = new a();
$obj->foo = 'bar';
$obj->bar = 1234;
$var = \Doctrine\Common\Util\Debug::export($obj, 2);
```

ostrolucky:
Fixed thanks to pretty interesting comment posted [12 years ago](http://php.net/manual/en/function.get-object-vars.php#47075)

ref: #386

### DIFF
--- a/lib/Doctrine/Common/Util/Debug.php
+++ b/lib/Doctrine/Common/Util/Debug.php
@@ -144,41 +144,26 @@ final class Debug
 
     /**
      * Fill the $return variable with class attributes
+     * Based on obj2array function from {@see http://php.net/manual/en/function.get-object-vars.php#47075}
      *
      * @param object   $var
-     * @param stdClass $return
+     * @param \stdClass $return
      * @param int      $maxDepth
      *
      * @return mixed
      */
     private static function fillReturnWithClassAttributes($var, \stdClass $return, $maxDepth)
     {
-        $reflClass = ClassUtils::newReflectionObject($var);
-        $parsedAttributes = array();
-        do {
-            $currentClassName = $reflClass->getName();
+        $clone = (array) $var;
 
-            foreach ($reflClass->getProperties() as $reflProperty) {
-                $attributeKey = $reflProperty->isPrivate() ? $currentClassName . '#' : '';
-                $attributeKey .= $reflProperty->getName();
-
-                if (isset($parsedAttributes[$attributeKey])) {
-                    continue;
-                }
-
-                $parsedAttributes[$attributeKey] = true;
-
-                $name =
-                      $reflProperty->getName()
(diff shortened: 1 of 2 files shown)


lib/Doctrine/Common/Util/Debug.php
```diff
@@ -144,41 +144,26 @@ public static function export($var, $maxDepth)
...
-                    . ($reflProperty->isPrivate() ? ':private' : '')
-                    . ($reflProperty->isProtected() ? ':protected' : '')
-                ;
-
-                $reflProperty->setAccessible(true);
-                $return->$name = self::export($reflProperty->getValue($var), $maxDepth - 1);
+        $clone = (array) $var;
+
+        foreach (array_keys($clone) as $key) {
+            $aux = explode("\0", $key);
```
Ocramius: Nice and clean!

lib/Doctrine/Common/Util/Debug.php
```diff
@@ -144,41 +144,26 @@ public static function export($var, $maxDepth)
...
-                $name =
-                      $reflProperty->getName()
-                    . ($return->__CLASS__ !== $currentClassName || $reflProperty->isPrivate() ? ':' . $currentClassName : '')
-                    . ($reflProperty->isPrivate() ? ':private' : '')
-                    . ($reflProperty->isProtected() ? ':protected' : '')
-                ;
-
-                $reflProperty->setAccessible(true);
-                $return->$name = self::export($reflProperty->getValue($var), $maxDepth - 1);
+        $clone = (array) $var;
```
Ocramius: One thing that this doesn't consider is that every property is now a reference. A simple way to fix this is to use `array_map()` with a recursive identity function.

See https://3v4l.org/Yhb9u

```php
<?php

$foo = 'bar';
$bar = ['foo' => & $foo];
$baz = (object) $bar;

$baz->foo = 'tab';

var_dump($bar);
```

ostrolucky: Cannot reproduce the problem, see new testcase

Ocramius: Ah, sorry, missed that this is just for output, so the references are not usable anyway. Sorry :-\

Ocramius:
Thanks @ostrolucky! Nice and clean :-)

PR #803, opened 2017-06-25
Fix #744 - Stop overwriting class docblock with '' triggered by ::class

### LINKED ISSUE #744: StaticReflectionParser is dumb, assumes Class::class references are class level
If I use the below PHP class structure, then parsing the annotations will always return an empty class level annotation.  The current HEAD of https://www.drupal.org/project/migrate_source_csv is broken right now using this pattern. It is a sneaky breakage, because Unit tests pass. It is only when I run integration tests that is breaks.

See https://www.drupal.org/node/2818891

``` php
/**
 * Source for CSV.
 *
 * @MigrateSource(
 *   id = "csv"
 * )
 */
class CSV extends SourcePluginBase {

  protected $fileClass = '';

  /**
   * {@inheritdoc}
   */
  public function __construct(array $configuration, $plugin_id, $plugin_definition, MigrationInterface $migration) {
    $this->fileClass = empty($configuration['file_class']) ? CSVFileObject::class : $configuration['file_class'];
  }
```

ostrolucky:


tests/Doctrine/Tests/Common/Reflection/ExampleAnnotationClass.php
```diff
@@ -9,6 +9,7 @@
  *   key = "value"
  * )
  */
+const foo = \stdClass::class;
```
Ocramius: Can we move this to a new class and a separate test, please?

lib/Doctrine/Common/Reflection/StaticReflectionParser.php
```diff
@@ -153,8 +153,10 @@ protected function parse()
                     $docComment = $token[1];
                     break;
                 case T_CLASS:
-                    $this->docComment['class'] = $docComment;
-                    $docComment = '';
+                    if ($docComment) {
```
Ocramius: Should we check for `'' !== $docComment` here?

ostrolucky: IMO this is more readable, because I don't want to replace docComment with any blank value

tests/Doctrine/Tests/Common/Reflection/ExampleAnnotationClass.php
ostrolucky: Not worth the effort, I will rather just remove this line. It was a bonus thing only.

lib/Doctrine/Common/Reflection/StaticReflectionParser.php
ostrolucky: IMO this is more readable, because Intent here is to stop replacing docComment with blank value. When there is strict check, reader of the code searches for reasons why does it allow some blank values and there is none.

PR #804, opened 2017-06-27
php7 fix

danionut90:


alcaeus:
@danionut90 please target 2.7 or master for the patch, 2.4 is no longer supported.

danionut90:
The bug is not present on versions > 2.4

Ocramius:
Closing as `invalid` then: unless it's a security issue, we don't patch older versions.

PR #805, opened 2017-07-06
[3.0][BC Break]  Require PHP 7.2, added parameter and return types, enable strict types

Majkl578:
* Migrated Doctrine\Common to PHP 7.2 - initial step for https://github.com/doctrine/doctrine2/issues/6529
* Added parameter types
* Added return types
* Enabled strict types

Requires at least PHP 7.2.0alpha3 (due to `object` type being used).

Excluded from migration:
- ClassLoader - deprecated, to be removed in 3.0
- Proxy - to be phased out for 3.0

Possible TODOs:
* [x] also add types for tests
* [ ] deprecate some mixed/union parameters in API
* [ ] document API changes
* [x] Fix PHPStan - blocked by https://github.com/phpstan/phpstan/issues/367

Tests are green. 💚
Travis is failing because it doesn't provide PHP 7.2 version yet - tracked in https://github.com/travis-ci/travis-ci/issues/7989.

```
$ php7.2-local vendor/bin/phpunit -v
PHPUnit 5.7.21 by Sebastian Bergmann and contributors.

Runtime:       PHP 7.2.0alpha3
Configuration: /www/doctrine/common/phpunit.xml.dist

...............................................................  63 / 246 ( 25%)
............................................................... 126 / 246 ( 51%)
............................................................... 189 / 246 ( 76%)
.........................................................       246 / 246 (100%)

Time: 129 ms, Memory: 8.00MB
```

Currently targets master, but should definitely go into develop branch (which doesn't exist yet - create one so we can retarget the PR 😎 ).

cc @Ocramius @lcobucci

### DIFF
--- a/lib/Doctrine/Common/Comparable.php
+++ b/lib/Doctrine/Common/Comparable.php
@@ -17,6 +17,8 @@
  * <http://www.doctrine-project.org>.
  */
 
+declare(strict_types=1);
+
 namespace Doctrine\Common;
 
 /**
--- a/lib/Doctrine/Common/Comparable.php
+++ b/lib/Doctrine/Common/Comparable.php
@@ -38,9 +38,9 @@ interface Comparable
      * This method should not check for identity using ===, only for semantical equality for example
      * when two different DateTime instances point to the exact same Date + TZ.
      *
-     * @param mixed $other
+     * @param object $other
      *
      * @return int
      */
-    public function compareTo($other);
+    public function compareTo(object $other): int;
 }
--- a/lib/Doctrine/Common/EventManager.php
+++ b/lib/Doctrine/Common/EventManager.php
@@ -17,6 +17,8 @@
  * <http://www.doctrine-project.org>.
  */
 
+declare(strict_types=1);
+
 namespace Doctrine\Common;
 
 /**
--- a/lib/Doctrine/Common/EventManager.php
+++ b/lib/Doctrine/Common/EventManager.php
@@ -50,7 +50,7 @@ class EventManager
      *
      * @return void
      */
-    public function dispatchEvent($eventName, EventArgs $eventArgs = null)
+    public function dispatchEvent(string $eventName, ?EventArgs $eventArgs = null): void
     {
         if (isset($this->_listeners[$eventName])) {
             $eventArgs = $eventArgs === null ? EventArgs::getEmptyInstance() : $eventArgs;
@@ -68,7 +68,7 @@ class EventManager
      *
      * @return array The event listeners for the specified event, or all event listeners.
      */
-    public function getListeners($event = null)
+    public function getListeners(?string $event = null): array
     {
         return $event ? $this->_listeners[$event] : $this->_listeners;
     }
@@ -80,7 +80,7 @@ class EventManager
      *
      * @return boolean TRUE if the specified event has any listeners, FALSE otherwise.
      */
-    public function hasListeners($event)
+    public function hasListeners(string $event): bool
     {
         return !empty($this->_listeners[$event]);
     }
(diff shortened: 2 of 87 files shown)


lib/Doctrine/Common/Comparable.php
```diff
@@ -38,9 +40,9 @@
      * This method should not check for identity using ===, only for semantical equality for example
      * when two different DateTime instances point to the exact same Date + TZ.
      *
-     * @param mixed $other
+     * @param object $other
      *
      * @return int
      */
-    public function compareTo($other);
+    public function compareTo(object $other): int;
```
Ocramius: All of these BC breaks need to be documented, together with a migration path :-\

lib/Doctrine/Common/EventManager.php
```diff
@@ -68,7 +70,7 @@ public function dispatchEvent($eventName, EventArgs $eventArgs = null)
      *
      * @return array The event listeners for the specified event, or all event listeners.
      */
-    public function getListeners($event = null)
+    public function getListeners(?string $event = null): array
```
PowerKiKi: shouldn't this read `string $event = null` without the `?` prefix ?

If the parameter has a default value of `null` anyway, is the `?` not redundant ?

alcaeus: No. `null` is a default value, and strictly speaking, it's an invalid default value for the type. These are valid:
1. `string $event` => must be string, must be given in function call
2. `string $event = ''` => must be string, can be omitted in which case an empty string is used
3. `?string $event` => must be string or null, must be given in function call
4. `?string $event = null` => must be string or null, can be omitted in which case `null` is used

`string $event = null` is invalid, as you'd expect a `string` but receive `null`.

alcaeus: The difference is that for one, the scalar type hint says "I want a string", while for the other it says "I want a string or null". The "feature" where adding a default type of `null` makes anything nullable should not overrule a concise scalar type declaration, which in this case is `?string`.

PowerKiKi: Fair enough, you might have convinced me :)

Majkl578: The fact that `string $event = null` works is sort of a flaw of PHP 7.0 where nullable types didn't exist and nullability was implied by default null value.
Anyhow, when I talk about this difference with other people, I usually use this example to show why explicit `?` matters (besides consistence):

```php
function foo(?int $bar = null) {} // accepts int or null
function bar(?int $arg = 123) {} // accepts int or null
```
vs.
```php
function foo(int $arg = null) {} // accepts int or null
function bar(int $arg = 123) {} // oops, accepts only int, no null here!
```
This is even more obvious in inheritance:
```php
class A {
    public function foo(int $arg = null) {}
}
class B extends A {
    public function foo(int $arg = 123) {} // this is an error, LSP violation
}
// PHP Warning:  Declaration of B::foo(int $arg = 123) should be compatible with A::foo(?int $arg = NULL)
```

PowerKiKi: You had me at "consistence", but the rest is a very good explanation. Thank you for taking the time to clarify this rule of thumb.

alcaeus: @Majkl578 great example, I'll be using that from now on. Thanks!

Majkl578:
Tests re-triggered with 7.2 on Travis, green.

lib/Doctrine/Common/Comparable.php
Majkl578: @Ocramius What do you suggest to do? Would it be ok to simply grep signature changes for now? Ithink creating an extended sheet of changes would be contraproductive ATM, since Common 3.0 might change a lot.

Majkl578:
Not applicable anymore, but served its purpose well back then. :wink:

PR #806, opened 2017-07-10
Fix #744 - Stop overwriting class docblock with '' triggered by ::class

### LINKED ISSUE #744: StaticReflectionParser is dumb, assumes Class::class references are class level

Sam152:


lib/Doctrine/Common/Reflection/StaticReflectionParser.php
```diff
@@ -153,8 +154,10 @@ protected function parse()
                     $docComment = $token[1];
                     break;
                 case T_CLASS:
-                    $this->docComment['class'] = $docComment;
-                    $docComment = '';
+                    if ($last_token !== T_PAAMAYIM_NEKUDOTAYIM) {
```
heddn: Would it make sense to add a comment to this value? It isn't really clear why or what the the value equates too.

Sam152: Personally I think the constant is a good enough indicator on it's own. It's pretty easy to lookup if you are unfamiliar with it.

Ocramius: See https://externals.io/message/99753#99770

tests/Doctrine/Tests/Common/Reflection/ExampleAnnotationClass.php
```diff
@@ -11,4 +11,11 @@
  */
 class ExampleAnnotationClass {
 
+  const foo = \stdClass::class;
+
+  /**
+   * Example with comment.
+   */
+  const bar = \stdClass::class;
```
Ocramius: @Sam152 these additional symbols should probably be in a separate class/test - try not modifying existing test assets, please

Sam152: I've added a param to the data provider for the existing annotation test and added a new class for the new test case.

Ocramius:
@Sam152 that looks much better, thanks!

lib/Doctrine/Common/Reflection/StaticReflectionParser.php
```diff
@@ -144,6 +144,7 @@ protected function parse()
         }
         $tokenParser = new TokenParser($contents);
         $docComment = '';
+        $last_token = FALSE;
```
Ocramius: Just a reminder: lowercase `false` please ;-)

I'm changing it on merge anyway.

Ocramius:
Merged, thanks @Sam152!

Ocramius:
Backported to `2.7` via doctrine/common@6b5b6a700eed1faf7eb76b4eae61b82d915ae6c8

Sam152:
Thanks @Ocramius!

PR #809, opened 2017-07-25
fixates dependencies to the last version that supports php 5.6

rnuernberg:


alcaeus:
Requiring PHP 7.1 was intentional. There's no need to switch back. Closing as `invalid`.

rnuernberg:
For every old version of doctrine? That is quite painful - may I ask why?

Ocramius:
Older versions will install just fine: composer will resolve the latest
compatible versions for you

rnuernberg:
@Ocramius ah sorry - my bad. My project accidentally used php 7.1 on my local machine. Once I switch it to php 5.6 it does indeed install correctly.

PR #812, opened 2017-08-01
[WIP] Add coding standard checks

TomasVotruba:
After bumping [min version to PHP 7.1 in Doctrine Project](http://www.doctrine-project.org/2017/07/25/php-7.1-requirement-and-composer.html) and successful setup of [coding standard for Nette](https://github.com/nette/coding-standard) to all its repositories, I'd like to help setup Doctrine Coding Standard that would help you to maintain code and review PRs.

[Doctrine\CodingStandard](https://github.com/doctrine/coding-standard) seems inactive and not used in projects and I want to change that.

In final version, all it will require for this repository add this line to `travis.yml`:

```yaml
vendor/bin/ecs check libs src --config vendor/doctrine/coding-standard/basic-standard.neon
```

**In this PR we only prepare rules setup and after you agree on them, I will send PR to Doctrine\CodingStandard that would be basically only repository to deal with coding standard changes, like [Nette\CodingStandard](http://github.com/nette/coding-standard).**

What do you think?

### Questions

1. ` ! ` does not comply with default PSR2 checkers from PHP-CS-Fixer and it changes it to `!`. I don't know a checker that would do that now, but it coud be done with custom fixer, yet it requires maintenance overhead that could make lead to ignoring the checker class later when bugs appear - do you really find it important?

2. What extra rules over PSR2 would you like include to?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,11 +13,14 @@ matrix:
   include:
     - php: 7.1
       env: PHPSTAN=1
+           CODING_STANDARD=1
 
 before_script:
   - composer --prefer-source install
   - if [[ $PHPSTAN = 1 ]]; then composer require --dev phpstan/phpstan:^0.7; fi
+  - if [[ $CODING_STANDARD = 1 ]]; then composer require --dev symplify/easy-coding-standard:^2.2; fi
 
 script:
   - ./vendor/bin/phpunit
   - if [[ $PHPSTAN = 1 ]]; then vendor/bin/phpstan analyse -c phpstan.neon -l 3 lib tests; fi
+  - if [[ $CODING_STANDARD = 1 ]]; then vendor/bin/ecs check lib; fi
--- a/easy-coding-standard.neon
+++ b/easy-coding-standard.neon
@@ -1,0 +1,10 @@
+# this will be placed in doctrine/coding-stanard, easy-accessible for anyone
+includes:
+    # PSR-2 rules: https://github.com/doctrine/coding-standard/blob/fe7696ba040327075c7bf1693c5d88bd64995090/ruleset.xml#L6
+    - vendor/symplify/easy-coding-standard/config/psr2-checkers.neon
+
+checkers:
+    # can be used instead of
+    # https://github.com/doctrine/coding-standard/blob/master/Sniffs/Strings/ConcatenationSpacingSniff.php
+    PhpCsFixer\Fixer\Operator\ConcatSpaceFixer:
+        spacing: one
--- a/lib/Doctrine/Common/ClassLoader.php
+++ b/lib/Doctrine/Common/ClassLoader.php
@@ -183,7 +183,7 @@ class ClassLoader
             return false;
         }
 
-        require ($this->includePath !== null ? $this->includePath . DIRECTORY_SEPARATOR : '')
+        require($this->includePath !== null ? $this->includePath . DIRECTORY_SEPARATOR : '')
                . str_replace($this->namespaceSeparator, DIRECTORY_SEPARATOR, $className)
                . $this->fileExtension;
 
@@ -200,7 +200,7 @@ class ClassLoader
      */
     public function canLoadClass($className)
     {
-        if ($this->namespace !== null && strpos($className, $this->namespace.$this->namespaceSeparator) !== 0) {
+        if ($this->namespace !== null && strpos($className, $this->namespace . $this->namespaceSeparator) !== 0) {
             return false;
         }
 
@@ -250,7 +250,7 @@ class ClassLoader
      */
     public static function getClassLoader($className)
     {
-         foreach (spl_autoload_functions() as $loader) {
+        foreach (spl_autoload_functions() as $loader) {
             if (is_array($loader)
                 && ($classLoader = reset($loader))
                 && $classLoader instanceof ClassLoader
--- a/lib/Doctrine/Common/EventArgs.php
+++ b/lib/Doctrine/Common/EventArgs.php
@@ -58,7 +58,7 @@ class EventArgs
      */
     public static function getEmptyInstance()
     {
-        if ( ! self::$_emptyEventArgsInstance) {
+        if (! self::$_emptyEventArgsInstance) {
             self::$_emptyEventArgsInstance = new EventArgs;
         }
 
--- a/lib/Doctrine/Common/Persistence/Event/PreUpdateEventArgs.php
+++ b/lib/Doctrine/Common/Persistence/Event/PreUpdateEventArgs.php
@@ -126,7 +126,7 @@ class PreUpdateEventArgs extends LifecycleEventArgs
      */
     private function assertValidField($field)
     {
-        if ( ! isset($this->entityChangeSet[$field])) {
+        if (! isset($this->entityChangeSet[$field])) {
             throw new \InvalidArgumentException(sprintf(
                 'Field "%s" is not a valid field of the entity "%s" in PreUpdateEventArgs.',
                 $field,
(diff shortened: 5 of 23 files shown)


Ocramius:
> ` ! ` does not comply with default PSR2 checkers from PHP-CS-Fixer and it changes it to !. I don't know a checker that would do that now, but it coud be done with custom fixer, yet it requires maintenance overhead that could make lead to ignoring the checker class later when bugs appear - do you really find it important?

Yes - that was one of the reasons to start the CS project

 > What extra rules over PSR2 would you like include to?

None at all for now

Note that we refrain from merging CS changes in general, as they bring little value, while they break everyone's open PRs, if there is (and there usually is) any long-standing issue

.travis.yml
```diff
@@ -13,11 +13,14 @@ matrix:
   include:
     - php: 7.1
       env: PHPSTAN=1
+           CODING_STANDARD=1
 
 before_script:
   - composer --prefer-source install
   - if [[ $PHPSTAN = 1 ]]; then composer require --dev phpstan/phpstan:^0.7; fi
+  - if [[ $CODING_STANDARD = 1 ]]; then composer require --dev symplify/easy-coding-standard:^2.2; fi
```
Ocramius: I understand that pushing other CS tools can be important, but I'd rather just use `./vendor/bin/phpcs`, which works very well out of the box. For me, that is the go-to solution, as otherwise we just add more work where things go wrong.

alcaeus:
Regarding the negation operator - if that's the only reason to keep a standard separate from PSR-2, would it be worth dropping the rule?

easy-coding-standard.neon
```diff
@@ -0,0 +1,10 @@
+# this will be placed in doctrine/coding-stanard, easy-accessible for anyone
```
Ocramius: Using the `.neon` syntax is something I really don't understand at all. IMO another reason to use PHPCS here...

TomasVotruba:
> Yes - that was one of the reasons to start the CS project

Ok, I'll prepare custom fixer then.

> None at all for now

:+1: 

>  Note that we refrain from merging CS changes in general, as they bring little value, while they break everyone's open PRs, if there is (and there usually is) any long-standing issue

I understand, it was same with Nette. Would it be fine to add setup to `.travis.yml` only for new PRs and not making tests fail?

Do you agree with the rest of proposal?

.travis.yml
TomasVotruba: I don't see a reason to use just one tool, since they have great ecosystem that work best together, see `nette/coding-standard.neon`: https://github.com/nette/coding-standard/blob/master/coding-standard-php56.neon

Instead of writing own sniff/fixer that is missing in the other tool (and burn out in the process), you can just use it.

You can always fallback to just one tool if this causes any troubles replacing this line.

easy-coding-standard.neon
TomasVotruba: Syntax is mostly the same with YAML.
NEON simply [allows class autocomplete](https://github.com/Symplify/EasyCodingStandard#dont-write-checker-classes-make-use-of-neon-plugin).

alcaeus: It's famously similar to YAML with the notable exception that it uses tabs instead of spaces for indentation. Class autocomplete is something the IDE offers, not a feature of the config file itself.

Ocramius:
> Would it be fine to add setup to `.travis.yml` only for new PRs and not making tests fail?

That would be a fabulous addition.

Overall, I'm fine with improvement, I'm just a bit skeptic about changing tools that are extremely stable and worked well for ages. Plus I personally rely on @jaapio's libs on a number of other packages :-)

easy-coding-standard.neon
Ocramius: That... kinda made it a dozen million times worse :-\

.travis.yml
Majkl578: I'm -1 on adding an unnecessary dependency on an _unknown_ project with questionable stability (some of us remember how @TomasVotruba's repositories magically disappeared without any deprecation/abandonment/notice from GitHub, breaking any and all depending libraries/code :/). Additionally it brings barely any added value over just using PHPCS directly (and a single custom rule would be stored in doctrine/coding-standard regardless so it wouldn't clutter other repositories). _</biased-opinion-based-on-experience>_

easy-coding-standard.neon
TomasVotruba: > That... kinda made it a dozen million times worse :-\

It's pain when you're used to it :)

TomasVotruba:
I don't care about the tool, just to make it actually run and use.

> @Ocramius That would be a fabulous addition.

Would this do? https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/master/src/Fixer/Operator/NotOperatorWithSuccessorSpaceFixer.php

lcobucci:
@TomasVotruba he was referring to https://github.com/squizlabs/PHP_CodeSniffer 😉

.travis.yml
```diff
@@ -13,11 +13,14 @@ matrix:
...
 
 script:
   - ./vendor/bin/phpunit
   - if [[ $PHPSTAN = 1 ]]; then vendor/bin/phpstan analyse -c phpstan.neon -l 3 lib tests; fi
+  - if [[ $CODING_STANDARD = 1 ]]; then vendor/bin/ecs check lib; fi
```
lcobucci: Could you also apply the build stages feature that Travis now offers? It makes things so much better (in terms of configuration and execution). You can use the ORM as an example: https://github.com/doctrine/doctrine2/blob/a0c0d3bf2a7f49020ad7e54d7d3319a251574230/.travis.yml#L26

TomasVotruba:
@lcobucci Are you sure? :) Why reinvent the wheel?

.travis.yml
TomasVotruba: Sure

lcobucci:
@TomasVotruba I'm sure, that's phpcs. The one you mentioned is phpcs-fixer which is completely different.

TomasVotruba:
@lcobucci I meant are you sure @Ocramius meants that repo? I never saw link to that and it doesn't make sense to me.

My point is the already exists, we can use that instead of writing the same one:
https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/f9abae03a6da058b6786142ae172c0747d3d3493/src/Fixer/Operator/NotOperatorWithSpaceFixer.php#L32

lcobucci:
@TomasVotruba that URL is [here](https://github.com/doctrine/coding-standard#doctrine-php_codesniffer-coding-standard). I mean, I'm fine with using `phpcs-fixer` but `phpcs` is the quite established tool we have.

TomasVotruba:
> @TomasVotruba that URL is here. I mean, I'm fine with using phpcs-fixer but phpcs is the most established tool we have.

@lcobucci That's why I suggested ECS, to prevent this picking opinonated "better tool". Both are powerful and both work best together. Most people prefer the one they already use without really trying the other one.

Nette [uses them both](https://github.com/nette/coding-standard/blob/2f935070b82fbe4b1da8e564a8dc6dcb9bbeca25/coding-standard-php56.neon#L13-L15) on all 15+ repositories with no issue (and you can have YAML based config, if you need).

lcobucci:
@TomasVotruba we've just released [a new version of doctrine/coding-standard](https://github.com/doctrine/coding-standard/releases/tag/1.0.0), which should make it easier to achieve what you're doing here. Thanks for your contribution 😉

TomasVotruba:
@lcobucci That's awesome :) 
Please ping me in PR when you put on Travis to Doctrine packages.

PR #815, opened 2017-08-30
Convert proxy factory auto generate mode to integer

dragosprotung:

PR #816, opened 2017-08-30
Convert proxy factory auto generate mode to integer

dragosprotung:
Clean PR based on #815

### DIFF
--- a/lib/Doctrine/Common/Proxy/AbstractProxyFactory.php
+++ b/lib/Doctrine/Common/Proxy/AbstractProxyFactory.php
@@ -69,6 +69,13 @@ abstract class AbstractProxyFactory
      */
     const AUTOGENERATE_EVAL = 3;
 
+    private const AUTOGENERATE_MODES = [
+        self::AUTOGENERATE_NEVER,
+        self::AUTOGENERATE_ALWAYS,
+        self::AUTOGENERATE_FILE_NOT_EXISTS,
+        self::AUTOGENERATE_EVAL,
+    ];
+
     /**
      * @var \Doctrine\Common\Persistence\Mapping\ClassMetadataFactory
      */
@@ -80,7 +87,7 @@ abstract class AbstractProxyFactory
     private $proxyGenerator;
 
     /**
-     * @var bool Whether to automatically (re)generate proxy classes.
+     * @var int Whether to automatically (re)generate proxy classes.
      */
     private $autoGenerate;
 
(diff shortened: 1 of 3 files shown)


lib/Doctrine/Common/Proxy/AbstractProxyFactory.php
```diff
@@ -69,6 +69,13 @@
      */
     const AUTOGENERATE_EVAL = 3;
 
+    private const AUTOGENERATE_MODES = [
+        self::AUTOGENERATE_NEVER,
+        self::AUTOGENERATE_ALWAYS,
+        self::AUTOGENERATE_FILE_NOT_EXISTS,
+        self::AUTOGENERATE_EVAL
```
Majkl578: There should be comma after last item.

lib/Doctrine/Common/Proxy/AbstractProxyFactory.php
```diff
@@ -93,12 +100,19 @@
...
+     *
+     * @throws \Doctrine\Common\Proxy\Exception\InvalidArgumentException When auto generate mode is not valid.
      */
     public function __construct(ProxyGenerator $proxyGenerator, ClassMetadataFactory $metadataFactory, $autoGenerate)
     {
         $this->proxyGenerator  = $proxyGenerator;
         $this->metadataFactory = $metadataFactory;
-        $this->autoGenerate    = (bool)$autoGenerate;
+
+        if ( ! in_array($autoGenerate, self::AUTOGENERATE_MODES, false)) {
```
Majkl578: You don't need false here as it's default behavior.

dragosprotung: It is the default value, but I wanted to be explicit that we allow other values

lcobucci: And why not casting it to int and making a strict comparison?

lcobucci:
@dragosprotung there are some CS violations, I've just pushed #817 to update our build process and verify coding standard violations. Do you want me to point out the issues or do you prefer to wait for it to be merged and just use `phpcs/phpcbf`?

lib/Doctrine/Common/Proxy/AbstractProxyFactory.php
dragosprotung: Normally I don't set properties until they are validated. But I guess in this case makes no difference as it's the constructor

dragosprotung:
@lcobucci I can see them from your PR, but I will wait for it to be merged if it will not take too long

Ocramius:
@lcobucci let's not pick on CS if a massive conflicting PR is coming in anyway :-P

lib/Doctrine/Common/Proxy/AbstractProxyFactory.php
```diff
@@ -69,6 +69,13 @@
      */
     const AUTOGENERATE_EVAL = 3;
 
+    private const AUTOGENERATE_MODES = [
```
Ocramius: YAY!

PR #819, opened 2017-09-12
doctrine/cache 1.7.1 does not support php 5.6

defser:
soctrine/cache increases the minimum supported PHP version to 7.1.0, This means that it is not possible for doctrine/common to require doctrine/cache 1.*

This must change in 1.6.*


Ocramius:
This is `invalid`: composer will just find a set of dependencies that match your requirements. No need to add constraints here.

PR #822, opened 2017-11-02
StaticReflectionClass PHP7.2 issue

alexpott:
Unfortunately PHP7.2 \ReflectionClass has slightly different semantics which means that \Doctrine\Common\Reflection\StaticReflectionClass has issues working on both PHP7.1 (and earlier) and PHP7.2.

The error you get on PHP7.2 is:
`Declaration of Doctrine\Common\Reflection\StaticReflectionClass::newInstance($args) should be compatible with ReflectionClass::newInstance(...$args)`

An earlier PR on this problem was https://github.com/doctrine/common/pull/775#pullrequestreview-21471640


alexpott:
The first commit is a tests only situation where we should see PHP7.1 pass but PHP7.2 fail.

Majkl578:
This looks like a PHP 7.2 BC break, have you reported it? https://3v4l.org/7igHL

alexpott:
@Majkl578 yep - it was reported in https://github.com/php/php-src/pull/2363#issuecomment-279246184 - I guess the comment there suggests that the rollback should have gone to master too.

Ocramius:
Relevant: https://github.com/Roave/BetterReflection/issues/317

lib/Doctrine/Common/Reflection/StaticReflectionClass.php
```diff
@@ -22,412 +22,417 @@
...
-    /**
-     * {@inheritDoc}
-     */
-    public function __toString()
-    {
-        throw new ReflectionException('Method not implemented');
+if (version_compare(phpversion(), '7.2', '<')) {
+    include __DIR__ . '/legacy/StaticReflectionClass.php';
+}
+else {
```
EclipseGc: Let's do the version check in a trait definition and just use that trait in this class to swap out the method.

alexpott:
Thanks @Ocramius that's super simple and appears to work for both 7.1 and 7.2. Yay!

dawehner:
@Ocramius You are a genius. Thank you!

Majkl578:
I'd rather wait for PHP to get this fixed on their side... There's still time to fix it before 7.2 goes stable.
There's already a bug report for this: https://bugs.php.net/bug.php?id=74292

EclipseGc:
But this won't work for 5.5 will it?

Ocramius:
5.5 is not supported anymore?

EclipseGc:
Right, I was making the point for Drupal which is still supporting 5.5 but would also like to support 7.2. The trait based approach (with an include for the 7.2 specific code) is the best solution for our needs I think.

alexpott:
So PHP have decided to rollback the BC break.

PR #823, opened 2017-11-21
Provide better check for aliased class name

### LINKED ISSUE #794: AbstractManagerRegistry::getManagerForClass doesn't work with anonymous classes
### Situation:

* PHP 7.1.5
* Using windows (it's an important part)
* Trying to use Symfony validator to validate anonymous class

The validator point is useless, but what's triggered behind is the<br>`$registry->getManagerForClass(get_class($myAnonymousClassInstance));`

### The test

```php
$a = new class {
    public $value;
};

var_dump(get_class($a));
```` 

Output on Windows:

`class@anonymous\x00E:\dev\www\skeleton\src\AppBundle\Controller\DefaultController.php000001982329027F` 

With this, I get an `ORMException`:

```
Unknown Entity namespace alias 'class@anonymousE'.
```

Obviously, because there is the `:` syntax, Doctrine is looking for a specific entity namespace.

---

I also tested this on linux inside a command and it works:

```php
$a = new class {
    public $value;
};

$manager = $this->getContainer()->get('doctrine')->getManagerForClass(get_class($a));

var_dump($class, $manager);
```

Output is correct because `$manager` is `null`:

```
"class@anonymous\x00/mnt/e/dev/www/skeleton/src/AppBundle/Command/TestCommand.php0x7f66b1aa5341"
null
```

### The fix

I think the `AbstractManagerRegistry::getManagerForClass()` should return `null` if the class name starts with `class@anonymous\x00`, whatever the platform, because it is the way PHP 7.1 handles anonymous classes names.

What do you think?

zlikavac32:
Hi,

So this is my idea on how to resolve #794 so any feedback is welcome

### DIFF
--- a/lib/Doctrine/Common/Persistence/AbstractManagerRegistry.php
+++ b/lib/Doctrine/Common/Persistence/AbstractManagerRegistry.php
@@ -19,6 +19,8 @@
 
 namespace Doctrine\Common\Persistence;
 
+use Doctrine\Common\Util\ClassUtils;
+
 /**
  * Abstract implementation of the ManagerRegistry contract.
  *
@@ -189,7 +191,7 @@ abstract class AbstractManagerRegistry implements ManagerRegistry
     public function getManagerForClass($class)
     {
         // Check for namespace alias
-        if (strpos($class, ':') !== false) {
+        if (ClassUtils::isClassNameAliasedClassName($class)) {
             list($namespaceAlias, $simpleClassName) = explode(':', $class, 2);
             $class                                  = $this->getAliasNamespace($namespaceAlias) . '\\' . $simpleClassName;
         }
--- a/lib/Doctrine/Common/Util/ClassUtils.php
+++ b/lib/Doctrine/Common/Util/ClassUtils.php
@@ -106,4 +106,20 @@ class ClassUtils
     {
         return rtrim($proxyNamespace, '\\') . '\\' . Proxy::MARKER . '\\' . ltrim($className, '\\');
     }
+
+    /**
+     * Checks whether provided class name is aliased class name or not.
+     *
+     * @param string $className
+     *
+     * @return bool
+     */
+    public static function isClassNameAliasedClassName($className)
+    {
+        if (strpos($className, ':') === false) {
+            return false;
+        }
+
+        return false === class_exists($className);
+    }
 }
--- a/tests/Doctrine/Tests/Common/Util/ClassUtilsTest.php
+++ b/tests/Doctrine/Tests/Common/Util/ClassUtilsTest.php
@@ -65,6 +65,26 @@ namespace Doctrine\Tests\Common\Util
             $reflClass = ClassUtils::newReflectionObject($object);
             self::assertEquals($expectedClassName, $reflClass->getName());
         }
+
+        /**
+         * @runInSeparateProcess
+         * @preserveGlobalState disabled
+         */
+        public function testExistingClassNameWithColonIsNotDetectedAsAlias()
+        {
+            class_alias(self::class, 'a:b');
+            self::assertFalse(ClassUtils::isClassNameAliasedClassName('a:b'));
+        }
+
+        public function testNonExistingClassNameWithColonIsDetectedAsAlias()
+        {
+            self::assertTrue(ClassUtils::isClassNameAliasedClassName('a:b'));
+        }
+
+        public function testClassNameWithoutColonIsNotDetectedAsAlias()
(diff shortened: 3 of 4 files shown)


lib/Doctrine/Common/Util/ClassUtils.php
```diff
@@ -106,4 +106,20 @@ public static function generateProxyClassName($className, $proxyNamespace)
...
+     *
+     * @return bool
+     */
+    public static function isClassNameAliasedClassName($className)
+    {
+        if (strpos($className, ':') === false) {
+            return false;
+        }
+
+        return false === class_exists($className);
```
Ocramius: This is not sufficient to define whether a class is aliased: `ClassUtils::isClassNameAliasedClassName('💩')` is unlikely a valid alias.

lib/Doctrine/Common/Persistence/AbstractManagerRegistry.php
```diff
@@ -189,7 +191,7 @@ public function getManager($name = null)
     public function getManagerForClass($class)
     {
         // Check for namespace alias
-        if (strpos($class, ':') !== false) {
```
Ocramius: We should really just deprecate class aliases: the `::class` meta-constant provided by PHP since PHP 5.5 is sufficient in most scenarios.

tests/Doctrine/Tests/Common/Util/ClassUtilsTest.php
```diff
@@ -65,6 +65,26 @@ public function testNewReflectionObject($className, $expectedClassName)
...
             self::assertEquals($expectedClassName, $reflClass->getName());
         }
+
+        /**
+         * @runInSeparateProcess
+         * @preserveGlobalState disabled
+         */
+        public function testExistingClassNameWithColonIsNotDetectedAsAlias()
+        {
+            class_alias(self::class, 'a:b');
```
Ocramius: This is a really bad idea - I don't expect the doctrine components to support this stuff explicitly :-\

lib/Doctrine/Common/Persistence/AbstractManagerRegistry.php
zlikavac32: That could work also. Should that be the focus now?

lib/Doctrine/Common/Util/ClassUtils.php
zlikavac32: It will not be detected as alias because it does not have a `:`. This PR's idea was to allow anonymous classes that have `:` in their FQN not to be treated as aliased class names.

Since I'm not that familiar with Doctrine internals, I tried to poke as little as possible and check whether a class name is real class (when there is `:` present) would mean that it's not aliased class

tests/Doctrine/Tests/Common/Util/ClassUtilsTest.php
zlikavac32: I may not follow you. This is here just to avoid PHP's engine dependent anonymous classes FQN's (which sometimes give `:` and sometimes don't) and to be able to test that case

Ocramius: Don't think we should support anything around `get_class()` with anonymous classes. If you want to go down that path,  reflecting + checking whether a class is anonymous would be better.

lib/Doctrine/Common/Persistence/AbstractManagerRegistry.php
Ocramius: I think so - the aliasing was really a pre-PHP-5.5 thing, because typing full class names was quite annoying.

zlikavac32: That would then be out of the scope of this issue and PR?

tests/Doctrine/Tests/Common/Util/ClassUtilsTest.php
zlikavac32: I agree, but this issue came out with Symfony validators that can validate arbitrary data, one of which may be anonymous classes. 

So it somehow made sense to report it here because right now, heuristic for aliased classes detects certain anonymous classes as aliased and because of that, exception is thrown.

To sum this up, it makes much more sense to go with the `::class` than to go with this heuristics?

Majkl578:
Hello,
Thanks.

zlikavac32:
I didn't see the notification. I'll check it out later and if it's still relevant, will create a new PR :).

Thank you

Issue #826, opened 2017-12-06
The future of Common 3.0


Majkl578:
The proposal is pretty simple: have no Common 3.0, instead split Common into smaller packages. We've already had discussion about this and Common 3.0 seems pretty much like relic and multiple components have already been split to their own packages.

Current proposal:

Persistence | Move to new package doctrine/persistence (rename namespace to Doctrine\Persistence).
-----|-----
Proxy | To be dropped without replacement.
Reflection | Not really sure about this one, either move to doctrine/reflection or drop it.
Util/ClassUtils | Proxy-related methods are obsolete makind the rest also pretty much obso.ete - drop it
Util/Debug | Drop it or move to doctrine/debug.
Util/Dumper | Unsure about its usefulness (there already exist more advanced tools like Tracy or Symfony Dumper),
Util/Inflector | Already moved out, only BC.
ClassLoader | Already deprecated, drop it.
CommonException | ~
Comparable | I know no place where Comparable is used across Doctrine (is there any?) but it seems useful i.e. for ORM 3.0 - move it to doctrine/comparable.
EventArgs + EventManager + EventSubscriber | We've already discussed replacing it by i.e. Symfony's EventDispatcher - drop it if suitable replacement is selected, doctrine/event-manager.
Lexer | Slready moved out, only BC.
NotifyPropertyChanged + PropertyChangedListener | These sound useful and are used in ORM so move it to doctrine/property-change-notifier (what a name!).
Version | ~

Each new package would use PHP 7.2 (or even 7.1 where 7.2 has no benefit), follow CS and would be ported to typed code.

WDYT?

malarzm:
>>> NotifyPropertyChanged + PropertyChangedListener These sound useful and are used in ORM so move it to doctrine/property-change-notifier
>>
>> Move to ORM, rework as isolated layer if possible

>Would ODM need these as well?

We'd need to port these as well

alcaeus:
> I know no place where Comparable is used across Doctrine (is there any?) but it seems useful i.e. for ORM 3.0 - move it to doctrine/comparable.

We were thinking of using it to compare value objects instead of comparing them by reference in ODM 2.0. Extracting it to a separate package would make sense.

alcaeus:
> The proposal is pretty simple: have no Common 3.0, instead split Common into smaller packages. 

[...]

PR #830, opened 2018-01-15
Use Null Coalesce Operator

carusogabriel:


Majkl578:
Not quite sure we need any further changes in doctrine/common ATM - this repository (and especially Proxy component) is near its EOL.

lcobucci:
@carusogabriel more info in #826

PR #832, opened 2018-02-16
Refactor generateConstructorImpl method

243083df:
Minor refactor.
Code seems more clean


Majkl578:
Hi, thanks for your PR.

We are going to phase out Proxy component soon so I'd rather prefer not to fiddle with its code unless it'd be a bugfix.
Since this change has no functional benefit, only slightly changes code structure and changes some whitespaces in the generated code, I am going to close this PR. Thank you. :)

PR #840, opened 2018-05-09
Fix CS

simPod:
I spotted this image

So I was wondering what if...a tiny contribution...might not hurt after all 🤔


Ocramius:
@simPod thanks!

PR #842, opened 2018-06-07
Separate Event Manager to doctrine/event-manager

Majkl578:
EventManager + EventSubscriber + EventArgs separated into [doctrine/event-manager](https://github.com/doctrine/event-manager).

One step closer to finishing #826.

_Note: `@dev` needs to be removed before 2.9.0 (once `1.0.0` is tagged)._

PR #843, opened 2018-06-11
Provide forward-compat replace for doctrine/{event-manager,persistence,reflection}

Majkl578:
This should allow us to depend on `doctrine/event-manager` package while still using `doctrine/common` 2.8.x (thus not requiring immediate release of 2.9, which should be done once we solve `doctrine/persistence`).
This is targetting 2.8 only, not applicable in 2.9.x-dev (#842 will be done instead).
Once `doctrine/common` 2.9.0 is out, real dependency kicks in.

@Ocramius Does this make sense as forward compatible polyfill? doctrine/event-manager is identical.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -23,6 +23,16 @@
     "require-dev": {
         "phpunit/phpunit": "^5.7"
     },
+    "replace": {
+        "doctrine/event-manager": "1.0.*",
+        "doctrine/persistence": "1.0.*",
+        "doctrine/reflection": "1.0.*"
+    },
+    "conflict": {
+        "doctrine/event-manager": ">=1.1.0@dev",
+        "doctrine/persistence": ">=1.1.0@dev",
+        "doctrine/reflection": ">=1.1.0@dev"
+    },
     "autoload": {
         "psr-4": {
             "Doctrine\\Common\\": "lib/Doctrine/Common"


Majkl578:
Hmm, not sure if this is gonna work. :thinking:

Majkl578:
Did some testing locally and it seems fine. :+1: Let's wait for @Ocramius to possibly spot some dragons.

Ocramius:
Hmm. @Majkl578, did you also check `"provides"`? Other than that, this seems to be the correct usage of `"replace"`, but `"replace"` shouldn't need `"conflict"`. This will likely need to be manually tested against a local satis installation before we push it and cause some issues downstream.

Majkl578:
I tried `provide` but it didn't seem to work as expected, maybe I just did something wrong.

> "replace" shouldn't need "conflict"

Hmm, that's to avoid 1.1-2.0 simultaneously which could, according to semver, contain some new features that Common wil no longer be able to provide.

> This will likely need to be manually tested against a local satis installation before we push it and cause some issues downstream.

Agreed, anyone volunteers for the job? 😊 
Alternatively we can _just_ ship Common 2.9.0 with no features, just separated EvM (and leave Persistence for 2.10).

stof:
@Majkl578 `replace` already means `provide`+`conflict`

Majkl578:
Updated to also include doctrine/persistence and doctrine/reflection.

2.9.0 will not include any of these three packages.

stof:
AFAIK, these 3 packages are ready. So I'd rather release them and move forward to 2.9.0 already, without messing up with `replace` rules here. It will be simpler.

Majkl578:
Yes, the plan is to release 2.9.0 in upcoming days (once #845 is resolved) and then, should there be some issues with upgrading (i.e. some combinations like DBAL 2.8 + ORM 2.6 + Symfony 3.x), we can consider this PR. For now I'll leave it open and we'll see.

Majkl578:
I think the transition in 2.9.0 went well and without any major issues, so this can now be closed.

PR #845, opened 2018-06-14
doctrine/common deprecations (soft)

Majkl578:
~_Note: Also includes commits from #844 - review only https://github.com/doctrine/common/pull/845/commits/1860aa60215b2aef3deb136de02d40783a113476 until merged._~

Deprecations for the rest of doctrine/common:
* whole Proxy namespace in favor of ocramius/proxy-manager
* Util\Debug in favor of symfony/var-dumper
* Util\ClassUtils without replacement (mostly an internal helper)
* Util\Inflector in favor of doctrine/inflector
* Lexer in favor of doctrine/lexer

### DIFF
--- a/lib/Doctrine/Common/Lexer.php
+++ b/lib/Doctrine/Common/Lexer.php
@@ -2,6 +2,10 @@
 namespace Doctrine\Common;
 
 use Doctrine\Common\Lexer\AbstractLexer;
+use function trigger_error;
+use const E_USER_DEPRECATED;
+
+@trigger_error(Lexer::class . ' is deprecated.', E_USER_DEPRECATED);
 
 /**
  * Base class for writing simple lexers, i.e. for creating small DSLs.
--- a/lib/Doctrine/Common/Lexer.php
+++ b/lib/Doctrine/Common/Lexer.php
@@ -13,6 +13,8 @@ use Doctrine\Common\Lexer\AbstractLexer;
  * @author Guilherme Blanco <guilhermeblanco@hotmail.com>
  * @author Jonathan Wage <jonwage@gmail.com>
  * @author Roman Borschel <roman@code-factory.org>
+ *
+ * @deprecated Use Doctrine\Common\Lexer\AbstractLexer from doctrine/lexer package instead.
  */
 abstract class Lexer extends AbstractLexer
 {
--- a/lib/Doctrine/Common/Util/ClassUtils.php
+++ b/lib/Doctrine/Common/Util/ClassUtils.php
@@ -9,6 +9,8 @@ use Doctrine\Common\Persistence\Proxy;
  *
  * @author Benjamin Eberlei <kontakt@beberlei.de>
  * @author Johannes Schmitt <schmittjoh@gmail.com>
+ *
+ * @deprecated The ClassUtils class is deprecated.
  */
 class ClassUtils
 {
--- a/composer.json
+++ b/composer.json
@@ -27,7 +27,8 @@
     "require-dev": {
         "phpunit/phpunit": "^6.3",
         "doctrine/coding-standard": "^1.0",
-        "squizlabs/php_codesniffer": "^3.0"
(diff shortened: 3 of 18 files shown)


alcaeus:
For classes that are part of a public API, I'd prefer if we use `@trigger_error(...)` to report these deprecations. That will make it easier for others to deal with these.

lib/Doctrine/Common/Lexer.php
```diff
@@ -13,6 +13,8 @@
  * @author Guilherme Blanco <guilhermeblanco@hotmail.com>
  * @author Jonathan Wage <jonwage@gmail.com>
  * @author Roman Borschel <roman@code-factory.org>
+ *
+ * @deprecated Use Doctrine\Common\Lexer\AbstractLexer from doctrine/lexer package instead.
```
Ocramius: Do we even want to have a `doctrine/lexer`? I thought the @hoa stuff would fix this, once their versioning system works too...

Majkl578: Yes we do. :)
https://github.com/doctrine/lexer

Majkl578:
@alcaeus What I'd like to do is ship soft deprecations now and then, after some months passed, ship trigger_errors. The issue is that we can't realistically trigger_error() for proxies ATM as it would unfixably break ORM.

alcaeus:
> The issue is that we can't realistically trigger_error() for proxies ATM as it would unfixably break ORM.

It would report deprecations for users that could easily be skipped if they properly configured their deprecation handler, which unfortunately seems to be the exception rather than the rule. So yeah, adding those `trigger_error` calls later works fine for me. 👍

lib/Doctrine/Common/Lexer.php
stof: We have it since 2013. We at least want people to use this one

stof:
I suggest we at least add the triggers already for things which should not be used anymore already (the BC classes for the lexer and inflector), as said in #846.

Reporting deprecations for the proxy system before the replacement is available is indeed a bad idea.
Thus, it would not really be about deprecating these classes but about marking doctrine/common as abandoned instead.

Majkl578:
@stof Makes sense. :+1: I added silent deprecations for those three.

lib/Doctrine/Common/Lexer.php
Ocramius: Alright :+1:

nicolas-grekas:
This PR breaks the Symfony test suite because there is no way to not trigger the deprecation while using ORM 2.7.x-dev. The ORM should be updated to use the not-deprecated code path when possible I suppose.

lib/Doctrine/Common/Util/ClassUtils.php
```diff
@@ -9,6 +9,8 @@
  *
  * @author Benjamin Eberlei <kontakt@beberlei.de>
  * @author Johannes Schmitt <schmittjoh@gmail.com>
+ *
+ * @deprecated The ClassUtils class is deprecated.
```
Soullivaneuh: The `ClassUtils` is useful to me. What would you suggest in replacement?

Soullivaneuh: Note: Was useful to resolve proxy classes and get the real class name.

Ocramius: This will no longer be applicable, because class resolution goes to the `ProxyManager` component for ORM 3.x

Soullivaneuh: @Ocramius So if I well understood, we have to keep the usage of this deprecated `ClassUtils` unti ORM v3 usage and then go to the ProxyManager class ?

Ocramius: Correct. In general, working with `get_class()` is discourage, but if you need it, you will probably have to use both classes during a transition period.

PR #849, opened 2018-07-22
Revert adding silent deprecations for legacy ClassLoader, Inflector and Lexer

Majkl578:
Reverts 39bc3964c485340b4cceba25e344a7e0d5f1ee4a in 2.9.

In #845 we added silent deprecations using the `@trigger_error(..., E_USER_DEPRECATED)` pattern (suggested in https://github.com/doctrine/common/pull/845#issuecomment-397631398).
This concept was originally created in a good faith to let users opt-in for deprecations only when they are interested in seeing them and aware it could happen (thus expecting deprecations may appear) and show none by default (for everyone else who doesn't opt-in).
Unfortunately Symfony had started misusing this concept and [has turned it to opt-out](https://github.com/symfony/symfony/issues/27936#issuecomment-404935867), just like if these were non-silent deprecations without `@` (messages logged, printed out and tests error out with exit code 1). In the end this means users are now seeing silenced deprecations unexpectedly without ever opting-in (in tests, debug toolbar etc.) and it breaks their test suite / CI.

Causes issues like: https://github.com/doctrine/doctrine2/issues/7306
Further (out-of-scope discussion): https://github.com/symfony/symfony/issues/27936

Proposal: Revert now for 2.9, release as 2.9.1. Then reevaluate usefulness of the concept of silent deprecations for future use.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -27,8 +27,7 @@
     "require-dev": {
         "phpunit/phpunit": "^6.3",
         "doctrine/coding-standard": "^1.0",
-        "squizlabs/php_codesniffer": "^3.0",
-        "symfony/phpunit-bridge": "^4.0.5"
+        "squizlabs/php_codesniffer": "^3.0"
     },
     "autoload": {
         "psr-4": {
--- a/lib/Doctrine/Common/ClassLoader.php
+++ b/lib/Doctrine/Common/ClassLoader.php
@@ -1,11 +1,6 @@
 <?php
 namespace Doctrine\Common;
 
-use function trigger_error;
-use const E_USER_DEPRECATED;
-
-@trigger_error(ClassLoader::class . ' is deprecated.', E_USER_DEPRECATED);
-
 /**
  * A <tt>ClassLoader</tt> is an autoloader for class files that can be
  * installed on the SPL autoload stack. It is a class loader that either loads only classes
--- a/lib/Doctrine/Common/Lexer.php
+++ b/lib/Doctrine/Common/Lexer.php
@@ -2,10 +2,6 @@
 namespace Doctrine\Common;
 
 use Doctrine\Common\Lexer\AbstractLexer;
-use function trigger_error;
-use const E_USER_DEPRECATED;
-
-@trigger_error(Lexer::class . ' is deprecated.', E_USER_DEPRECATED);
 
 /**
  * Base class for writing simple lexers, i.e. for creating small DSLs.
--- a/lib/Doctrine/Common/Util/Inflector.php
+++ b/lib/Doctrine/Common/Util/Inflector.php
@@ -2,10 +2,6 @@
 namespace Doctrine\Common\Util;
 
 use Doctrine\Common\Inflector\Inflector as BaseInflector;
-use function trigger_error;
-use const E_USER_DEPRECATED;
-
-@trigger_error(Inflector::class . ' is deprecated.', E_USER_DEPRECATED);
 
 /**
  * Doctrine inflector has static methods for inflecting text.
--- a/tests/Doctrine/Tests/Common/ClassLoaderTest.php
+++ b/tests/Doctrine/Tests/Common/ClassLoaderTest.php
@@ -4,9 +4,6 @@ namespace Doctrine\Tests\Common;
 
 use Doctrine\Common\ClassLoader;
 
-/**
- * @group legacy
- */
 class ClassLoaderTest extends \Doctrine\Tests\DoctrineTestCase
 {
     public function testClassLoader()


Ocramius:
> Proposal: Revert now for 2.9, release as 2.9.1. Then reevaluate usefulness of the concept of silent deprecations for future use.

I don't understand this bit: this would lead to a patch release with functional changes, no?

stof:
Well, I would rather make a 2.6.3 release of the ORM which would have the fix merged in it to make the ORM stop using these deprecated APIs. The goal of these warnings is precisely to make it clear when a deprecated API is used, to push to migrate. The ORM was still using some APIs deprecated years ago.

The phpunit-bridge opting-in for the report of deprecations has precisely this goal: reporting when your code uses deprecated APIs.
And I would not say that Symfony turned it into an opt-out. The phpunit-bridge is not a mandatory requirement. If you don't want to opt-in for deprecation reporting in your tests, don't install it.

nicolas-grekas:
This is not the right move for Doctrine and its users (including Symfony): triggering silenced deprecations is the way to go. As @stof explained above and on https://github.com/symfony/symfony/issues/27936, depractions *are* opt-in on Symfony also.

Majkl578:
> triggering silenced deprecations is the way to go.

It would be if Symfony hadn't started turning them into hard errors for _anything_ in vendor without intentionally enabling this behavior.
Package A calling deprecated API of package B should not make user's test suite error out, that is not acceptable.

> As @stof explained above and on symfony/symfony#27936, depractions are opt-in on Symfony also.

No, they are not, see https://github.com/symfony/symfony/issues/27936#issuecomment-404935867. This erroneous behavior is enabled by default and exits with code 1.

nicolas-grekas:
The right move IMHO is to resubmit https://github.com/symfony/symfony/pull/27609 now that Doctrine doesn't use its own deprecated things internally.

Majkl578:
To make it clear what the expected behavior is (to me and also to other end-user consumers):
1) I *do* want reporting when my code in src/ calls _any_ deprecated API from src/
2) I *do* want reporting when my code in src/ calls deprecated API from vendor/ *directly*
3) I *do not* want reporting when my code in src/ calls non-deprecated API in vendor/ that _indirectly_ calls some deprecated API of another package in vendor/ (out of scope of my app)
4) I *do not* want reporting when some package in vendor/ calls deprecated API of another package in vendor/ (out of scope of my app)

Symfony's default behavior currently violates 3) and 4).

At this moment I recommend abandoning the concept of `@trigger_error(...)` and using https://github.com/phpstan/phpstan-deprecation-rules instead.

> The right move IMHO is to resubmit symfony/symfony#27609 now that Doctrine doesn't use its own deprecated things internally.

Agreed that it should be resubmitted, but this PR is rather about issues like https://github.com/doctrine/doctrine2/issues/7306 because of https://github.com/doctrine/doctrine2/pull/7307.

nicolas-grekas:
I'd recommend to close this PR.
On Symfony's side, tests are now green with dev branches of doctrine/*
About the defaults, things will be fixed once https://github.com/symfony/recipes/pull/442 is merged.
These deprecations will help ppl moving to v3. By adding, we allow ppl to more easily notice when their code will break *before* it actually happens. Reciprocally, by removing these deprecations, some people will notice the BC break is a hard way: by triggering fatal errors after moving to v3. This would be much worse (not everyone on Earth runs static analyses, nor even has 100% test coverage.)

Majkl578:
The default behavior of PHPUnit Bridge is still wrong in my eyes, by _default_ it's reporting issues that are out-of-scope for the _project_ where it's used.
Just like you don't run static analysis, CS or tests on `vendor/` directory, PHPUnit bridge should not report deprecations that happen within `vendor/` - In other words, _3)_ and _4)_ from my [previous comment](https://github.com/doctrine/common/pull/849#issuecomment-407053196) are still happening.

> On Symfony's side, tests are now green with dev branches of `doctrine/*`.

This is primarily about _everyone out there_ who now gets deprecation errors and test failures just because Doctrine ORM internally calls deprecated ClassLoader from Doctrine Common.

alcaeus:
> The default behavior of PHPUnit Bridge is still wrong in my eyes, by `default` it's reporting issues that are out-of-scope for the `project` where it's used.

That's not our problem, and that's not a reason to drop these deprecation warnings. The change to the phpunit-bridge prevents failing builds due to these deprecations while still reporting them (which people can turn off using `weak_vendors` if they want). If people complain about these deprecations, we tell them why they're not relevant to them and how they can ignore them. For another solution, see next point.

> This is primarily about `everyone out there` who now gets deprecation errors and test failures just because Doctrine ORM internally calls deprecated ClassLoader from Doctrine Common.

The fix to that is publishing a release with these usages dropped and be done with it, not get into an even lengthier discussion about these deprecations. Obviously, the deprecations will show up for some users who aren't running the latest version, but if you're running the latest version of one dependency and an older version of another that's something you have to expect and deal with. After the deprecation is done, the focus should be to drop the deprecated behavior as soon as possible.

As for reevaluating "usefulness of the concept of silent deprecations for future use": these deprecations are extremely useful when preparing a major version upgrade, so we should keep them in at all costs. If the tooling around them is sub-optimal, that's not our problem. Instead, this should be fixed upstream instead of dropping the concept altogether. With that said, I'm closing this PR.

PR #851, opened 2018-08-25
Run tests on PHP 7.3

mnapoli:


mnapoli:
Ah nevermind sorry for the noise! I thought this was available (because the build showed up correctly on another project) but it isn't yet.

Ocramius:
Will be merged when ready :+1:

mnapoli:
For reference: https://travis-ci.org/CouscousPHP/Couscous/jobs/420526698

I was mistaken because the build appears in Travis so I thought "great!", but then the build failed…

mnapoli:
Thanks for the link!

PR #861, opened 2018-12-04
[DDC-2556] Fix generating non-loading proxy methods when the identifier field is defined in a trait

Jalle19:
I stumbled upon this issue today while dealing with a nasty performance issue. After some digging around in the various bug trackers I found https://github.com/doctrine/doctrine2/issues/3282 which illustrates the issue, however it doesn't make it clear what the root cause is.

The problem is that the code that gets checked against the regular expression when determining if a getter is "a short identifier getter" is not the right code. With this patch the actual method in the trait gets checked, not the various use statements in the class that uses said trait.

In my particular case, the lines that got considered where these:

```
use XXX\ContentApi\Domain\Traits\HasLikes;
use XXX\ContentApi\Domain\Traits\Identifiable;
use XXX\ContentApi\Infrastructure\Traits\AutoIncrements;
use XXX\ContentApi\Infrastructure\Traits\Timestamps;
```

Here, the line numbers are correct, but the file is wrong, which ultimately causes the check to fail.

### DIFF
--- a/tests/Doctrine/Tests/Common/Proxy/ProxyLogicIdentifierGetterTest.php
+++ b/tests/Doctrine/Tests/Common/Proxy/ProxyLogicIdentifierGetterTest.php
@@ -78,6 +78,13 @@ class ProxyLogicIdentifierGetterTest extends \PHPUnit\Framework\TestCase
             ]
         );
 
+        $methods = array_merge(
+            $methods,
+            [
+                [new LazyLoadableObjectWithTraitClassMetadata(), 'identifierFieldInTrait', 123],
+            ]
+        );
+
         return array_merge(
             $methods,
             [
--- a/tests/Doctrine/Tests/Common/Proxy/ProxyLogicIdentifierGetterTest.php
+++ b/tests/Doctrine/Tests/Common/Proxy/ProxyLogicIdentifierGetterTest.php
@@ -57,40 +57,18 @@ class ProxyLogicIdentifierGetterTest extends \PHPUnit\Framework\TestCase
      */
     public function methodsForWhichLazyLoadingShouldBeDisabled()
     {
-        $methods = [
+        return [
             [new LazyLoadableObjectClassMetadata(), 'protectedIdentifierField', 'foo'],
+            [new LazyLoadableObjectWithTypehintsClassMetadata(), 'identifierFieldNoReturnTypehint', 'noTypeHint'],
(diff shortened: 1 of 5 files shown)


tests/Doctrine/Tests/Common/Proxy/LazyLoadableObjectWithTypehints.php
```diff
@@ -12,6 +12,7 @@
  */
 class LazyLoadableObjectWithTypehints
 {
+    use IdentifierFieldTrait;
```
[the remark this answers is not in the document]
Jalle19: Done

Jalle19:
> @Jalle19 yeah, I think the change can make it into `doctrine/common` in the next minor, since the change is trivial and clear

This would be more than great, that way I wouldn't have to potentially resort to pulling in my own fork just for this fix.

tests/Doctrine/Tests/Common/Proxy/ProxyLogicIdentifierGetterTest.php
```diff
@@ -78,6 +78,13 @@ public function methodsForWhichLazyLoadingShouldBeDisabled()
             ]
         );
 
+        $methods = array_merge(
+            $methods,
+            [
+                [new LazyLoadableObjectWithTraitClassMetadata(), 'identifierFieldInTrait', 123],
```
Ocramius: Why isn't this just in the `array_merge()` block below?

Jalle19: Basically I followed the same style since there was already a separate block of methods merged in. I can change them all to just be one big array.

Jalle19: Current style seems to be a relic from c0c18190c8e3220e70e279f2c927465a834fa706

Ocramius: Since we don't support PHP <7.1 anymore, this can be simplified now :+1:

Jalle19: Fixed

Jalle19:
No problem, thanks for reviewing and merging so quickly!

PR #862, opened 2018-12-09
fix method reference in @see doblock

vladyslavstartsev:


Ocramius:
@vladyslavstartsev thanks!

Issue #868, opened 2019-04-08
Bug: Proxy: Support of PHP 7 return type hints in magic method


janbarasek:
Hello,

I am using version `v2.10.0` with PHP `7.2.14`.

In entities I use trait `SmartObject` (https://doc.nette.org/en/2.4/smartobject) implements magic methods (for example `__isset()`) with scalar type hints.

I know new version of `ProxyGenerator` make proxy methods with type hints, but in magic methods not.

On runtime I get a PHP error: `Declaration of ShopUp\Database\Proxy\__CG__\ShopUp\UserAdvance\Entity\AdvanceUser::__isset($name) must be compatible with ShopUp\Security\Entity\User::__isset(string $name): bool`

Generated proxy class:

<img width="746" alt="Snímek obrazovky 2019-04-08 v 10 27 39" src="https://user-images.githubusercontent.com/4738758/55709584-f0dfe300-59e8-11e9-9795-acfd026484b9.png">

But in real entity is:

<img width="661" alt="Snímek obrazovky 2019-04-08 v 10 28 08" src="https://user-images.githubusercontent.com/4738758/55709611-01905900-59e9-11e9-9072-c056bca76a00.png">

Thank you for support.

janbarasek:
Please merge my changes and create new version: https://github.com/doctrine/common/pull/869

rybajz:
Hi,
I have the same problem. Is there any progress? Can you merge @janbarasek suggest?

Thanks

janbarasek:
@rybajz I understand your trouble but I I could not make a correction that meets all the requirements.

Typical case when is problem is `SmartObject` trait in Nette Framework: https://doc.nette.org/en/3.0/smartobject

platx:
Any updates? I got the same issue with some external lib, that couldn't be changed by me, but i should use proxy functional for this lib.

malarzm:
It seems like https://github.com/doctrine/common/pull/869 still has some not addressed comments. Do you want to pick it up and finish?

janbarasek:
Finally done! And we're going to celebrate! :)

rybajz:
Great!

PR #869, opened 2019-04-08
Fix proxies with return types in magic methods

janbarasek:
On PHP version 7.0 or newest automatically add scalar return type hints to magic methods for compatibility generated Proxy class with real entity.

More information: https://github.com/doctrine/common/issues/868

### DIFF
--- a/lib/Doctrine/Common/Proxy/ProxyGenerator.php
+++ b/lib/Doctrine/Common/Proxy/ProxyGenerator.php
@@ -1,10 +1,10 @@
 <?php
 namespace Doctrine\Common\Proxy;
 
-use Doctrine\Persistence\Mapping\ClassMetadata;
 use Doctrine\Common\Proxy\Exception\InvalidArgumentException;
 use Doctrine\Common\Proxy\Exception\UnexpectedValueException;
 use Doctrine\Common\Util\ClassUtils;
+use Doctrine\Persistence\Mapping\ClassMetadata;
 use function array_map;
 use function method_exists;
 
@@ -433,14 +433,24 @@ EOT;
         $hasParentGet         = false;
         $returnReference      = '';
         $inheritDoc           = '';
+        $name                 = '$name';
+        $parametersString     = '$name';
+        $returnTypeHint       = null;
 
         if ($reflectionClass->hasMethod('__get')) {
-            $hasParentGet = true;
-            $inheritDoc   = '{@inheritDoc}';
+            $hasParentGet     = true;
+            $inheritDoc       = '{@inheritDoc}';
+            $methodReflection = $reflectionClass->getMethod('__get');
 
-            if ($reflectionClass->getMethod('__get')->returnsReference()) {
+            if ($methodReflection->returnsReference()) {
                 $returnReference = '& ';
             }
+
+            $methodParameters = $methodReflection->getParameters();
+            $name             = '$' . $methodParameters[0]->getName();
+
+            $parametersString = $this->buildParametersString($methodReflection->getParameters());
+            $returnTypeHint   = $this->getMethodReturnType($methodReflection);
         }
 
         if (empty($lazyPublicProperties) && ! $hasParentGet) {
@@ -450,19 +460,28 @@ EOT;
         $magicGet = <<<EOT
     /**
      * $inheritDoc
-     * @param string \$name
+     * @param string $name
      */
-    public function {$returnReference}__get(\$name)
+    public function {$returnReference}__get($parametersString)$returnTypeHint
     {
 
 EOT;
 
         if ( ! empty($lazyPublicProperties)) {
-            $magicGet .= <<<'EOT'
-        if (\array_key_exists($name, self::$lazyPropertiesNames)) {
-            $this->__initializer__ && $this->__initializer__->__invoke($this, '__get', [$name]);
+            $magicGet .= sprintf(<<<'EOT'
+        if (\array_key_exists(%s, self::$lazyPropertiesNames)) {
+            $this->__initializer__ && $this->__initializer__->__invoke($this, '__get', [%s]);
+EOT
+                , $name, $name);
 
-            return $this->$name;
+            if ($returnTypeHint === ': void') {
+                $magicGet .= "\n            return;";
+            } else {
+                $magicGet .= "\n            return \$this->$name;";
+            }
+
+            $magicGet .= <<<'EOT'
+
         }
 
 
@@ -470,22 +489,34 @@ EOT;
         }
 
         if ($hasParentGet) {
-            $magicGet .= <<<'EOT'
-        $this->__initializer__ && $this->__initializer__->__invoke($this, '__get', [$name]);
+            $magicGet .= sprintf(<<<'EOT'
+        $this->__initializer__ && $this->__initializer__->__invoke($this, '__get', [%s]);
+EOT
+                , $name);
 
-        return parent::__get($name);
+            if ($returnTypeHint === ': void') {
+                $magicGet .= sprintf(<<<'EOT'
 
-EOT;
+        parent::__get(%s);
+        return;
+EOT
+                    , $name);
+            } else {
+                $magicGet .= sprintf(<<<'EOT'
+
+        return parent::__get(%s);
+EOT
+                    , $name);
+            }
         } else {
-            $magicGet .= <<<'EOT'
-        trigger_error(sprintf('Undefined property: %s::$%s', __CLASS__, $name), E_USER_NOTICE);
+            $magicGet .= sprintf(<<<EOT
+        trigger_error(sprintf('Undefined property: %%s::$%%s', __CLASS__, %s), E_USER_NOTICE);
 
-EOT;
+EOT
+                , $name);
         }
 
-        $magicGet .= "    }";
-
-        return $magicGet;
+        return $magicGet . "\n    }";
     }
 
     /**
@@ -499,22 +530,31 @@ EOT;
     {
         $lazyPublicProperties = $this->getLazyLoadedPublicPropertiesNames($class);
         $hasParentSet         = $class->getReflectionClass()->hasMethod('__set');
+        $parametersString     = '$name, $value';
+        $returnTypeHint       = null;
+
+        if ($hasParentSet) {
+            $methodReflection = $class->getReflectionClass()->getMethod('__set');
+            $parametersString = $this->buildParametersString($methodReflection->getParameters());
+            $returnTypeHint   = $this->getMethodReturnType($methodReflection);
+        }
 
         if (empty($lazyPublicProperties) && ! $hasParentSet) {
             return '';
         }
 
         $inheritDoc = $hasParentSet ? '{@inheritDoc}' : '';
-        $magicSet   = <<<EOT
+        $magicSet   = sprintf(<<<'EOT'
     /**
-     * $inheritDoc
-     * @param string \$name
-     * @param mixed  \$value
+     * %s
+     * @param string $name
+     * @param mixed  $value
      */
-    public function __set(\$name, \$value)
+    public function __set(%s)%s
     {
 
-EOT;
+EOT
+            , $inheritDoc, $parametersString, $returnTypeHint);
 
         if ( ! empty($lazyPublicProperties)) {
             $magicSet .= <<<'EOT'
@@ -540,9 +580,7 @@ EOT;
             $magicSet .= "        \$this->\$name = \$value;";
         }
 
-        $magicSet .= "\n    }";
-
-        return $magicSet;
+        return $magicSet . "\n    }";
     }
 
     /**
@@ -556,6 +594,14 @@ EOT;
     {
         $lazyPublicProperties = $this->getLazyLoadedPublicPropertiesNames($class);
         $hasParentIsset       = $class->getReflectionClass()->hasMethod('__isset');
+        $parametersString     = '$name';
+        $returnTypeHint       = null;
+
+        if ($hasParentIsset) {
+            $methodReflection = $class->getReflectionClass()->getMethod('__isset');
+            $parametersString = $this->buildParametersString($methodReflection->getParameters());
+            $returnTypeHint   = $this->getMethodReturnType($methodReflection);
+        }
 
         if (empty($lazyPublicProperties) && ! $hasParentIsset) {
             return '';
@@ -568,7 +614,7 @@ EOT;
      * @param  string \$name
      * @return boolean
      */
-    public function __isset(\$name)
+    public function __isset($parametersString)$returnTypeHint
     {
 
 EOT;
@@ -590,7 +636,6 @@ EOT;
         $this->__initializer__ && $this->__initializer__->__invoke($this, '__isset', [$name]);
 
         return parent::__isset($name);
-
 EOT;
         } else {
             $magicIsset .= "        return false;";
--- a/phpstan.neon.dist
+++ b/phpstan.neon.dist
@@ -37,6 +37,9 @@ parameters:
         -
             message: '#^Property Doctrine\\Tests\\Common\\Proxy\\ProxyLogicVoidReturnTypeTest::\$initializerCallbackMock \(callable&PHPUnit\\Framework\\MockObject\\MockObject\) does not accept PHPUnit\\Framework\\MockObject\\MockObject&stdClass\.$#'
             path: '%currentWorkingDirectory%/tests/Doctrine/Tests/Common/Proxy/ProxyLogicVoidReturnTypeTest.php'
+        -
+            message: '#^Method Doctrine\\Tests\\Common\\Proxy\\MagicIssetClassWithInteger::__isset\(\) should return bool but returns int\.$#'
+            path: '%currentWorkingDirectory%/tests/Doctrine/Tests/Common/Proxy/MagicIssetClassWithInteger.php'
 
 includes:
     - vendor/phpstan/phpstan-phpunit/extension.neon
--- a/tests/Doctrine/Tests/Common/Proxy/MagicGetClassWithScalarType.php
+++ b/tests/Doctrine/Tests/Common/Proxy/MagicGetClassWithScalarType.php
@@ -1,0 +1,39 @@
+<?php
+
+namespace Doctrine\Tests\Common\Proxy;
+
+/**
+ * Test asset class
+ * @author Jan Barasek <jan@barasek.com>
+ */
+class MagicGetClassWithScalarType
+{
+    /**
+     * @var string
+     */
+    public $id = 'id';
+
+    /**
+     * @var string
+     */
+    public $publicField = 'publicField';
+
+    /**
+     * @param string $name
+     *
+     * @return string
+     * @throws \BadMethodCallException
+     */
+    public function __get(string $name): string
+    {
+        if ($name === 'test') {
+            return 'test';
+        }
+
+        if ($name === 'publicField' || $name === 'id') {
+            throw new \BadMethodCallException('Should never be called for "publicField" or "id"');
+        }
+
+        return 'not defined';
+    }
+}
--- a/tests/Doctrine/Tests/Common/Proxy/MagicGetClassWithScalarTypeAndRenamedParameter.php
+++ b/tests/Doctrine/Tests/Common/Proxy/MagicGetClassWithScalarTypeAndRenamedParameter.php
@@ -1,0 +1,39 @@
+<?php
+
+namespace Doctrine\Tests\Common\Proxy;
+
+/**
+ * Test asset class
+ * @author Jan Barasek <jan@barasek.com>
+ */
+class MagicGetClassWithScalarTypeAndRenamedParameter
+{
+    /**
+     * @var string
+     */
+    public $id = 'id';
+
+    /**
+     * @var string
+     */
+    public $publicField = 'publicField';
+
+    /**
+     * @param string $n
+     *
+     * @return string
+     * @throws \BadMethodCallException
+     */
+    public function __get(string $n): string
+    {
+        if ($n === 'test') {
+            return 'test';
+        }
+
+        if ($n === 'publicField' || $n === 'id') {
+            throw new \BadMethodCallException('Should never be called for "publicField" or "id"');
+        }
+
+        return 'not defined';
+    }
+}
--- a/tests/Doctrine/Tests/Common/Proxy/MagicGetClassWithVoid.php
+++ b/tests/Doctrine/Tests/Common/Proxy/MagicGetClassWithVoid.php
@@ -1,0 +1,21 @@
+<?php
+
+namespace Doctrine\Tests\Common\Proxy;
+
+/**
+ * Magic class asset test with void return type for getter
+ */
+class MagicGetClassWithVoid
+{
+
+    /**
+     * @param string $name
+     *
+     * @return void
+     * @throws \BadMethodCallException
+     */
+    public function __get(string $name): void
+    {
+        return;
+    }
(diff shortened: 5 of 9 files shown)


Ocramius:
These additions require careful unit and integration testing

janbarasek:
> These additions require careful unit and integration testing

@Ocramius I think it's ok, because add type hints to `__isset()` method is safe. If I add type and extended method don't use types, it's will works. Or no?

lib/Doctrine/Common/Proxy/ProxyGenerator.php
```diff
@@ -548,7 +574,19 @@ private function generateMagicIsset(ClassMetadata $class)
...
         $inheritDoc = $hasParentIsset ? '{@inheritDoc}' : '';
-        $magicIsset = <<<EOT
+        if (PHP_VERSION_ID >= 70000){
+			$magicIsset = <<<EOT
+    /**
+     * $inheritDoc
+     * @param  string \$name
+     * @return boolean
+     */
+    public function __isset(string \$name): bool
```
alcaeus: This is no good: there's nothing preventing me from using a different return type hint other than `bool`: https://3v4l.org/8gHmF. You should instead read the type hint from the original method and use that.

janbarasek: @alcaeus ok, I understand. Thank you for feedback.

New version check real parameter type hint and return type by \ReflectionClass from original entity.

Is it ok? https://github.com/doctrine/common/pull/869/commits/a3b6ed150737ab1725f64a719127adbb2cb2097c

lib/Doctrine/Common/Proxy/ProxyGenerator.php
```diff
@@ -982,6 +968,41 @@ private function getParameterType(\ReflectionParameter $parameter)
...
+	 *
+	 * @return string|null
+	 */
+	private function getMethodParameterTypeHintByReflection(\ReflectionClass $reflection, string $method, string $parameter)
+	{
+		try {
+			$typeHint = $this->getParameterTypeByMethod($reflection->getMethod($method), $parameter);
+
+			return $typeHint ? $typeHint . ' ' : null;
+		} catch (\ReflectionException $e) {
```
Ocramius: This should *not* happen

lib/Doctrine/Common/Proxy/ProxyGenerator.php
```diff
@@ -574,29 +565,24 @@ private function generateMagicIsset(ClassMetadata $class)
...
-    public function __isset(string \$name): bool
-    {
+		$nameParameterTypeHint = $this->getMethodParameterTypeHintByReflection($class->getReflectionClass(), '__isset', 'name');
 
-EOT;
-		} else {
-			$magicIsset = <<<EOT
+		try {
+			$returnTypeHint = $this->getMethodReturnType($class->getReflectionClass()->getMethod('__isset'));
+		} catch (\ReflectionException $e) {
```
Ocramius: This should *not* happen

lib/Doctrine/Common/Proxy/ProxyGenerator.php
```diff
@@ -433,27 +433,23 @@ private function generateMagicGet(ClassMetadata $class)
...
-    public function {$returnReference}__get(string \$name)
-    {
+		$nameParameterTypeHint = $this->getMethodParameterTypeHintByReflection($class->getReflectionClass(), '__get', 'name');
 
-EOT;
-		} else{
-			$magicGet = <<<EOT
+		try {
+			$returnTypeHint = $this->getMethodReturnType($class->getReflectionClass()->getMethod('__get'));
+		} catch (\ReflectionException $e) {
```
Ocramius: This should *not* happen

lib/Doctrine/Common/Proxy/ProxyGenerator.php
janbarasek: Snímek obrazovky 2019-04-08 v 15 31 16

Can you explain me why? My IDE say I have a unhandled exception.

Thanks.

Ocramius: `getMethod()` can fail if the method does not exist. Silencing the failure is not a viable solution if that happens though: if you can't handle it, re-throw instead.

janbarasek: Fixed.

lib/Doctrine/Common/Proxy/ProxyGenerator.php
janbarasek: Catch section removed.

lib/Doctrine/Common/Proxy/ProxyGenerator.php
janbarasek: Catch section removed.

janbarasek:
I don't know how to fix this error: https://travis-ci.org/doctrine/common/jobs/517424516

In last commits I add new test classes represent `: bool`, `: int` and `: void` cases.

I think reliably `void` return type hint detection is not possible. Source: https://stackoverflow.com/questions/1809764/check-if-php-function-returns-null-or-nothing

How can I fix this error? Remove `return`?

Snímek obrazovky 2019-04-08 v 21 19 22

Thank you.

janbarasek:
@Ocramius I fixed all tests: https://travis-ci.org/doctrine/common/builds/517742787

But last test failed in compile time: https://travis-ci.org/doctrine/common/jobs/517742791

I think I can not solve it. Can you accept my change? Missing typehints breaks all my web applications.

Thank you.

lib/Doctrine/Common/Proxy/ProxyGenerator.php
```diff
@@ -427,18 +428,24 @@ private function generateMagicGet(ClassMetadata $class)
             if ($reflectionClass->getMethod('__get')->returnsReference()) {
                 $returnReference = '& ';
             }
-        }
+
+			$nameParameterTypeHint = $this->getMethodParameterTypeHintByReflection($reflectionClass, '__get', 'name');
+			$returnTypeHint = $this->getMethodReturnType($reflectionClass->getMethod('__get'));
+        } else {
+			$nameParameterTypeHint = null;
```
Ocramius: Please set these to a default before the `if` block: `else` block can then be omitted

lib/Doctrine/Common/Proxy/ProxyGenerator.php
```diff
@@ -427,18 +428,24 @@ private function generateMagicGet(ClassMetadata $class)
...
+			$returnTypeHint = null;
+		}
 
         if (empty($lazyPublicProperties) && ! $hasParentGet) {
             return '';
         }
 
-        $magicGet = <<<EOT
+		$magicGet = <<<EOT
```
Ocramius: You changed `    ` for `\t` all over the patch: please revert that

lib/Doctrine/Common/Proxy/ProxyGenerator.php
```diff
@@ -480,24 +487,36 @@ public function {$returnReference}__get(\$name)
...
-        $hasParentSet         = $class->getReflectionClass()->hasMethod('__set');
+		$reflectionClass      = $class->getReflectionClass();
+        $hasParentSet         = $reflectionClass->hasMethod('__set');
+
+		if ($hasParentSet) {
+			$nameParameterTypeHint = $this->getMethodParameterTypeHintByReflection($reflectionClass, '__set', 'name');
+			$valueParameterTypeHint = $this->getMethodParameterTypeHintByReflection($reflectionClass, '__set', 'value');
+			$returnTypeHint = $this->getMethodReturnType($reflectionClass->getMethod('__set'));
+		} else {
+			$nameParameterTypeHint = null;
```
Ocramius: Same as above: please use a default value instead of an `else` block

lib/Doctrine/Common/Proxy/ProxyGenerator.php
```diff
@@ -537,24 +556,35 @@ public function __set(\$name, \$value)
...
         $lazyPublicProperties = array_keys($this->getLazyLoadedPublicProperties($class));
-        $hasParentIsset       = $class->getReflectionClass()->hasMethod('__isset');
+		$reflectionClass      = $class->getReflectionClass();
+        $hasParentIsset       = $reflectionClass->hasMethod('__isset');
+
+		if ($hasParentIsset) {
+			$nameParameterTypeHint = $this->getMethodParameterTypeHintByReflection($reflectionClass, '__isset', 'name');
+			$returnTypeHint = $this->getMethodReturnType($reflectionClass->getMethod('__isset'));
+		} else {
+			$nameParameterTypeHint = null;
```
Ocramius: Same as above

lib/Doctrine/Common/Proxy/ProxyGenerator.php
```diff
@@ -427,18 +428,24 @@ private function generateMagicGet(ClassMetadata $class)
             if ($reflectionClass->getMethod('__get')->returnsReference()) {
                 $returnReference = '& ';
             }
-        }
+
+			$nameParameterTypeHint = $this->getMethodParameterTypeHintByReflection($reflectionClass, '__get', 'name');
```
Ocramius: What happens if the parent class parameter name is not called `'name'`?

For instance:

```php
public function __get(string $n);
```

Consider using positional indexes for this API instead: that should rule out this scenario.

lib/Doctrine/Common/Proxy/ProxyGenerator.php
```diff
@@ -943,6 +973,38 @@ private function getParameterType(\ReflectionParameter $parameter)
         return $this->formatType($parameter->getType(), $parameter->getDeclaringFunction(), $parameter);
     }
 
+	/**
+	 * @param \ReflectionMethod $method
+	 * @param string $parameterName
+	 *
+	 * @return string|null
+	 */
+    private function getParameterTypeByMethod(\ReflectionMethod $method, string $parameterName)
```
Ocramius: As mentioned above, we should use positional parameter indexes instead: that removes ambiguity between proxy naming and original class naming

lib/Doctrine/Common/Proxy/ProxyGenerator.php
```diff
@@ -943,6 +973,38 @@ private function getParameterType(\ReflectionParameter $parameter)
...
+
+	/**
+	 * @param \ReflectionClass $reflection
+	 * @param string $method
+	 * @param string $parameter
+	 *
+	 * @return string|null
+	 * @throws \ReflectionException
+	 */
+	private function getMethodParameterTypeHintByReflection(\ReflectionClass $reflection, string $method, string $parameter)
```
Ocramius: Please declare return types on all affected/added methods

lib/Doctrine/Common/Proxy/ProxyGenerator.php
Ocramius: Note: this needs an added test case

lib/Doctrine/Common/Proxy/ProxyGenerator.php
Ocramius: Assets can be skipped in the static analysis steps

lib/Doctrine/Common/Proxy/ProxyGenerator.php
janbarasek: Case `__get(string $n)` added and works.

lib/Doctrine/Common/Proxy/ProxyGenerator.php
janbarasek: @Ocramius Added skipping.

lib/Doctrine/Common/Proxy/ProxyGenerator.php
```diff
@@ -480,11 +493,22 @@ public function {$returnReference}__get(\$name)
...
     private function generateMagicSet(ClassMetadata $class)
     {
         $lazyPublicProperties = $this->getLazyLoadedPublicProperties($class);
-        $hasParentSet         = $class->getReflectionClass()->hasMethod('__set');
+        $reflectionClass      = $class->getReflectionClass();
+        $hasParentSet         = $reflectionClass->hasMethod('__set');
+        $parametersBuffer     = '$name, $value';
+        $returnTypeHint       = null;
+
+
```
alcaeus:
```suggestion
```

lib/Doctrine/Common/Proxy/ProxyGenerator.php
```diff
@@ -419,14 +420,26 @@ private function generateMagicGet(ClassMetadata $class)
...
+            if ($methodReflection->returnsReference()) {
                 $returnReference = '& ';
             }
+
+            $methodParameters = $methodReflection->getParameters();
+            if (isset($methodParameters[0])) {
+                $firstParameterName = '$' . $methodParameters[0]->getName();
+            }
+
+            $parametersBuffer = $this->getMethodParametersWithTypes($methodReflection);
```
alcaeus: You should be able to use `buildParametersString` for this.

lib/Doctrine/Common/Proxy/ProxyGenerator.php
```diff
@@ -480,11 +493,22 @@ public function {$returnReference}__get(\$name)
...
-        $hasParentSet         = $class->getReflectionClass()->hasMethod('__set');
+        $reflectionClass      = $class->getReflectionClass();
+        $hasParentSet         = $reflectionClass->hasMethod('__set');
+        $parametersBuffer     = '$name, $value';
+        $returnTypeHint       = null;
+
+
+        if ($hasParentSet) {
+            $methodReflection = $reflectionClass->getMethod('__set');
+            $parametersBuffer = $this->getMethodParametersWithTypes($methodReflection);
```
alcaeus: `buildParametersString`

lib/Doctrine/Common/Proxy/ProxyGenerator.php
```diff
@@ -537,24 +561,35 @@ public function __set(\$name, \$value)
...
         $lazyPublicProperties = array_keys($this->getLazyLoadedPublicProperties($class));
-        $hasParentIsset       = $class->getReflectionClass()->hasMethod('__isset');
+        $reflectionClass      = $class->getReflectionClass();
+        $hasParentIsset       = $reflectionClass->hasMethod('__isset');
+        $parametersBuffer     = '$name';
+        $returnTypeHint       = null;
+
+        if ($hasParentIsset) {
+            $methodReflection = $reflectionClass->getMethod('__isset');
+            $parametersBuffer = $this->getMethodParametersWithTypes($methodReflection);
```
alcaeus: `buildParametersString`

lib/Doctrine/Common/Proxy/ProxyGenerator.php
janbarasek: @alcaeus Fixed, I am very sorry.

lib/Doctrine/Common/Proxy/ProxyGenerator.php
janbarasek: Replaced.

lib/Doctrine/Common/Proxy/ProxyGenerator.php
janbarasek: Replaced.

lib/Doctrine/Common/Proxy/ProxyGenerator.php
```diff
@@ -436,35 +449,35 @@ private function generateMagicGet(ClassMetadata $class)
...
+    public function {$returnReference}__get($parametersBuffer)$returnTypeHint
     {
 
 EOT;
 
         if ( ! empty($lazyPublicProperties)) {
-            $magicGet .= <<<'EOT'
-        if (array_key_exists($name, $this->__getLazyProperties())) {
-            $this->__initializer__ && $this->__initializer__->__invoke($this, '__get', [$name]);
+            $magicGet .= <<<EOT
```
Majkl578: Please keep NOWDOC and use `%s` + `sprintf()` instead - using escaping and variables inside HEREDOC makes the code even harder to read.

lib/Doctrine/Common/Proxy/ProxyGenerator.php
```diff
@@ -436,35 +449,35 @@ private function generateMagicGet(ClassMetadata $class)
...
         }
 
 
 EOT;
         }
 
         if ($hasParentGet) {
-            $magicGet .= <<<'EOT'
-        $this->__initializer__ && $this->__initializer__->__invoke($this, '__get', [$name]);
+            $magicGet .= <<<EOT
```
Majkl578: Please keep NOWDOC and use `%s` + `sprintf()` instead - using escaping and variables inside HEREDOC makes the code even harder to read.

lib/Doctrine/Common/Proxy/ProxyGenerator.php
```diff
@@ -436,35 +449,35 @@ private function generateMagicGet(ClassMetadata $class)
...
+        \$this->__initializer__ && \$this->__initializer__->__invoke(\$this, '__get', [$firstParameterName]);
 
-        return parent::__get($name);
+        return parent::__get($firstParameterName);
 
 EOT;
         } else {
-            $magicGet .= <<<'EOT'
```
Majkl578: Please keep NOWDOC and use `%s` + `sprintf()` instead - using escaping and variables inside HEREDOC makes the code even harder to read.

lib/Doctrine/Common/Proxy/ProxyGenerator.php
```diff
@@ -419,14 +420,26 @@ private function generateMagicGet(ClassMetadata $class)
...
+            $inheritDoc       = '{@inheritDoc}';
+            $methodReflection = $reflectionClass->getMethod('__get');
 
-            if ($reflectionClass->getMethod('__get')->returnsReference()) {
+            if ($methodReflection->returnsReference()) {
                 $returnReference = '& ';
             }
+
+            $methodParameters = $methodReflection->getParameters();
+            if (isset($methodParameters[0])) {
```
Majkl578: This will be always-true, please remove the condition.

lib/Doctrine/Common/Proxy/ProxyGenerator.php
```diff
@@ -419,14 +420,26 @@ private function generateMagicGet(ClassMetadata $class)
         $hasParentGet         = false;
         $returnReference      = '';
         $inheritDoc           = '';
+        $parametersBuffer     = '$name';
+        $firstParameterName   = '$name';
```
Majkl578: No need to call this "first", `__get()` can always has exactly one parameter.

lib/Doctrine/Common/Proxy/ProxyGenerator.php
```diff
@@ -419,14 +420,26 @@ private function generateMagicGet(ClassMetadata $class)
         $hasParentGet         = false;
         $returnReference      = '';
         $inheritDoc           = '';
+        $parametersBuffer     = '$name';
```
Majkl578: Please name this `$parametersString`.

lib/Doctrine/Common/Proxy/ProxyGenerator.php
```diff
@@ -411,6 +411,7 @@ public function __construct($initializer = null, $cloner = null)
      * @param \Doctrine\Common\Persistence\Mapping\ClassMetadata $class
      *
      * @return string
+     * @throws \ReflectionException
```
Majkl578: Please remove, besides being private API only, this can never happen as we check for `hasMethod()` before calling `getMethod()`.

lib/Doctrine/Common/Proxy/ProxyGenerator.php
```diff
@@ -480,11 +493,21 @@ public function {$returnReference}__get(\$name)
...
```
Majkl578: Please remove, besides being private API only, this can never happen as we check for `hasMethod()` before calling `getMethod()`.

lib/Doctrine/Common/Proxy/ProxyGenerator.php
```diff
@@ -537,24 +560,35 @@ public function __set(\$name, \$value)
...
```
Majkl578: Please remove, besides being private API only, this can never happen as we check for `hasMethod()` before calling `getMethod()`.

lib/Doctrine/Common/Proxy/ProxyGenerator.php
```diff
@@ -480,11 +493,21 @@ public function {$returnReference}__get(\$name)
...
      */
     private function generateMagicSet(ClassMetadata $class)
     {
         $lazyPublicProperties = $this->getLazyLoadedPublicProperties($class);
-        $hasParentSet         = $class->getReflectionClass()->hasMethod('__set');
+        $reflectionClass      = $class->getReflectionClass();
+        $hasParentSet         = $reflectionClass->hasMethod('__set');
+        $parametersBuffer     = '$name, $value';
```
Majkl578: Please name this `$parametersString`.

lib/Doctrine/Common/Proxy/ProxyGenerator.php
```diff
@@ -537,24 +560,35 @@ public function __set(\$name, \$value)
...
      * @return string
+     * @throws \ReflectionException
      */
     private function generateMagicIsset(ClassMetadata $class)
     {
         $lazyPublicProperties = array_keys($this->getLazyLoadedPublicProperties($class));
-        $hasParentIsset       = $class->getReflectionClass()->hasMethod('__isset');
+        $reflectionClass      = $class->getReflectionClass();
+        $hasParentIsset       = $reflectionClass->hasMethod('__isset');
+        $parametersBuffer     = '$name';
```
Majkl578: Please name this `$parametersString`.

lib/Doctrine/Common/Proxy/ProxyGenerator.php
```diff
@@ -497,7 +520,7 @@ private function generateMagicSet(ClassMetadata $class)
      * @param string \$name
      * @param mixed  \$value
```
Majkl578: In case of `__get()` we generate correct inherited parameter names. We should either do the same here or not do it for `__get()` but we should stay consistent.

I would personally generate only `@inheritdoc` when parent exists, for all magic methods.

lib/Doctrine/Common/Proxy/ProxyGenerator.php
janbarasek: I think this is correct. Can you send me how to fix this issue?

lib/Doctrine/Common/Proxy/ProxyGenerator.php
janbarasek: Fixed.

janbarasek:
Sorry, I don't know how to fix the tests.

Can you help me, please?

miloshavlicek:
@janbarasek I've sent you a PR https://github.com/janbarasek/common/pull/2 ... it should be fixed.

miloshavlicek:
I work on some fixes, so please wait for me.

alcaeus:
Thanks @janbarasek and @miloshavlicek for getting this fixed!

flk-a:
Where is error at genarated proxy file, in setter:
`    public function __set($name, $val)`
should be 
`    public function __set($name, $value)`

miloshavlicek:
@flk-a Have you tried this patch? https://github.com/doctrine/common/pull/897

miloshavlicek:
> > @flk-a Have you tried this patch? #897
> 
> I've received patch 2.13.2 by composer automatically. So, big amont of data was lost before i found this error.

It's already fixed, but not yet in 2.13.x branch, please see #897

PR #873, opened 2019-06-22
Use getName() instead __toString()

### LINKED ISSUE #872: PHP 7.4 Function ReflectionType::__toString() is deprecated
We have 
https://github.com/doctrine/common/blob/6bc33d1f26fb8653cda552e8e69f15d1e8512ade/lib/Doctrine/Common/Proxy/ProxyGenerator.php#L1024

Is it safe change it to `$name = $type->getName();`?

esbobkov:
Fix #872
P.S 
Didn't understand about nullability 
https://github.com/doctrine/common/issues/872#issuecomment-50442107

### DIFF
--- a/lib/Doctrine/Common/Proxy/ProxyGenerator.php
+++ b/lib/Doctrine/Common/Proxy/ProxyGenerator.php
@@ -1021,7 +1021,7 @@ EOT;
         \ReflectionMethod $method,
         \ReflectionParameter $parameter = null
     ) {
-        $name      = (string) $type;
+        $name      = $type->getName();
         $nameLower = strtolower($name);
 
         if ('self' === $nameLower) {


lib/Doctrine/Common/Proxy/ProxyGenerator.php
```diff
@@ -1021,7 +1021,7 @@ private function formatType(
         \ReflectionMethod $method,
         \ReflectionParameter $parameter = null
     ) {
-        $name      = (string) $type;
+        $name      = $type->getName();
```
Ocramius: The generated string for a nullable type must be `?\Foo`, not just `\Foo`

esbobkov: I think this part handle nullable already https://github.com/doctrine/common/blob/93479a34af73b55a7ac085315c62ca73b135ceeb/lib/Doctrine/Common/Proxy/ProxyGenerator.php#L1054-L1058

Ocramius: Hmm, yep, patch seems good as-is 👍

Ocramius:
Thanks @esbobkov!

PR #874, opened 2019-07-01
Fallback to 'copy' and 'unlink' instead of 'rename'

klammbueddel:
...to move a file without side effects.

This gist reproduces the problem:
https://gist.github.com/alcaeus/a367e895e6c55f7fb93870dcba46efa9

Same issue is fixed in a similary way in symfony core:
https://github.com/symfony/symfony/issues/12533

resolves #6713


klammbueddel:
closed due to alternative dev setup as suggested in #327

PR #875, opened 2019-09-10
Require PHP 7.4 tests to pass

alcaeus:
This updates PHPUnit to the latest version to ensure tests don't fail on PHP 7.4 (see https://github.com/sebastianbergmann/phpunit/pull/3765). 7.4 build is now required to pass as we're in the RC stage.

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -741,27 +741,29 @@
         },
         {
             "name": "doctrine/instantiator",
-            "version": "1.1.0",
+            "version": "1.2.0",
             "source": {
                 "type": "git",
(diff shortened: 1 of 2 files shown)


composer.lock
```diff
@@ -741,27 +741,29 @@
         },
         {
             "name": "doctrine/instantiator",
-            "version": "1.1.0",
```
alcaeus: The entire diff is the result of `cu --with-dependencies phpunit/phpunit` - I didn't want to update any other dependencies.

PR #877, opened 2019-10-28
ProxyGenerator: Update regex pattern to match simple id getters with …

simPod:
- [ ]  add tests
- [x]  test on production ready project

Currently, in ProxyGenerator there's support for generating simple (not triggering entity hydration) id getters in proxy class only for `return $this->identifier` 

```php
public function getId() { return $this->id; }
```
However, I also needed to support

```php
public function getId() { return MyId::new($this->id); }
```
and
```php
public function getId() { return new MyId($this->id); }
```

Would it make sense?

### DIFF
--- a/lib/Doctrine/Common/Proxy/ProxyGenerator.php
+++ b/lib/Doctrine/Common/Proxy/ProxyGenerator.php
@@ -21,7 +21,7 @@ class ProxyGenerator
      * Used to match very simple id methods that don't need
      * to be decorated since the identifier is known.
      */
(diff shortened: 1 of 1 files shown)


Ocramius:
This is already solved in ORM `master`, where a different lazy-loading strategy is used. Sadly, inspecting code via regex is too fragile: let's avoid that.

simPod:
Agree regex is not the proper tool, I'm heading to try master

simPod:
Cannot try it, doctrine bundle seems not to be adopted yet. Thanks for info though!

simPod:
I had to solve it anyway. It IMO improves current state. Feel free to close if not applicable.

Ocramius:
Since this is already solved in `master`, and we don't improve on this component anymore, I'd rather focus efforts in getting current `master` state shipped

PR #878, opened 2019-10-28
ProxyGenerator: Update regex pattern to match simple id getters with …

simPod:
_Followup, couldn't reopen https://github.com/doctrine/common/pull/877_  
_I had to solve it anyway. It IMO improves current state. Feel free to close if not applicable._

- [x]  add tests
- [x]  test on production ready project

Currently, in ProxyGenerator there's support for generating simple (not triggering entity hydration) id getters in proxy class only for `return $this->identifier` 

```php
public function getId() { return $this->id; }
```
However, I also needed to support

```php
public function getId() { return MyId::new($this->id); }
```
and
```php
public function getId() { return new MyId($this->id); }
```

### DIFF
--- a/lib/Doctrine/Common/Proxy/ProxyGenerator.php
+++ b/lib/Doctrine/Common/Proxy/ProxyGenerator.php
@@ -7,6 +7,8 @@ use Doctrine\Common\Util\ClassUtils;
 use Doctrine\Persistence\Mapping\ClassMetadata;
 use function array_map;
 use function method_exists;
+use function preg_match;
+use function sprintf;
 
 /**
  * This factory is used to generate proxy classes.
@@ -20,8 +22,19 @@ class ProxyGenerator
     /**
      * Used to match very simple id methods that don't need
      * to be decorated since the identifier is known.
+     *
+     * @see https://regex101.com/r/u72479/2/
      */
-    const PATTERN_MATCH_ID_METHOD = '((public\s+)?(function\s+%s\s*\(\)\s*)\s*(?::\s*\??\s*\\\\?[a-z_\x7f-\xff][\w\x7f-\xff]*(?:\\\\[a-z_\x7f-\xff][\w\x7f-\xff]*)*\s*)?{\s*return\s*\$this->%s;\s*})i';
+    const PATTERN_MATCH_ID_METHOD = <<<REGEXP
+~(
+(public\s+)?
+(function\s+%s\s*\(\)\s*)\s*
+(?::\s*\??\s*((?:\\\\?[\w\\x7f-\\xff][\w\\x7f-\\xff]*)+)\s*)?
+{\s*
+return\s*(\\\$this->%2\$s|new\s+\g{4}\(\\\$this->%2\$s\)|\g{4}::\w+\(\\\$this->%2\$s\));
+\s*}
+)~x
+REGEXP;
 
     /**
      * The namespace that contains all proxy classes.
--- a/tests/Doctrine/Tests/Common/Proxy/LazyLoadableObjectWithCustomIdType.php
+++ b/tests/Doctrine/Tests/Common/Proxy/LazyLoadableObjectWithCustomIdType.php
@@ -1,0 +1,26 @@
+<?php
+namespace Doctrine\Tests\Common\Proxy;
+
+use Doctrine;
+
+/**
+ * Test asset representing a lazy loadable object
+ */
+class LazyLoadableObjectWithCustomIdType
+{
+    /** @var string */
+    private $identifierFieldWithStaticVOConstructor;
+
+    /** @var string */
+    private $identifierFieldWithVOConstructor;
+
+    public function getIdentifierFieldWithStaticVOConstructor() : ValueId
+    {
+        return ValueId::new($this->identifierFieldWithStaticVOConstructor);
+    }
+
+    public function getIdentifierFieldWithVOConstructor() : ValueId
+    {
+        return new ValueId($this->identifierFieldWithVOConstructor);
+    }
+}
--- a/tests/Doctrine/Tests/Common/Proxy/LazyLoadableObjectWithCustomIdTypeClassMetadata.php
+++ b/tests/Doctrine/Tests/Common/Proxy/LazyLoadableObjectWithCustomIdTypeClassMetadata.php
@@ -1,0 +1,164 @@
+<?php
+namespace Doctrine\Tests\Common\Proxy;
+
+use Doctrine\Persistence\Mapping\ClassMetadata;
+use ReflectionClass;
+
+/**
+ * Class metadata test asset for @see LazyLoadableObject
+ */
+class LazyLoadableObjectWithCustomIdTypeClassMetadata implements ClassMetadata
+{
+    /**
+     * @var ReflectionClass
+     */
+    protected $reflectionClass;
+
+    /**
+     * @var array
+     */
+    protected $identifier = [
+        'identifierFieldWithStaticVOConstructor' => true,
+        'identifierFieldWithVOConstructor' => true,
+    ];
+
+    /**
+     * @var array
+     */
+    protected $fields = [
+        'identifierFieldWithStaticVOConstructor' => true,
+        'identifierFieldWithVOConstructor' => true,
+    ];
+
+    /**
+     * {@inheritDoc}
+     */
+    public function getName()
+    {
+        return $this->getReflectionClass()->getName();
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    public function getIdentifier()
+    {
+        return array_keys($this->identifier);
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    public function getReflectionClass()
+    {
+        if (null === $this->reflectionClass) {
+            $this->reflectionClass = new \ReflectionClass(__NAMESPACE__ . '\LazyLoadableObjectWithCustomIdType');
+        }
+
+        return $this->reflectionClass;
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    public function isIdentifier($fieldName)
+    {
+        return isset($this->identifier[$fieldName]);
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    public function hasField($fieldName)
+    {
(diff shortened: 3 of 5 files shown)


malarzm:
@simPod never actually tried that, but would it be possible to create a custom DBAL type that would deal with conversion? That way you wouldn't need to create the object on your own thus proxy wouldn't require adjustments

simPod:
@malarzm yes it would but we decided not to use custom dbal types if possible. There's huge configuration overhead to map each value object and we're trying to avoid as much configuration as possible.

What we do is we store primitives or very common values (uuid) in orm entity properties and convert them inside entity methods from/to value objects. The approach works very well combined with phpstan.

alcaeus:
TBH, I don't see this getting merged: the entirety of what's left in doctrine/common is deprecated, especially these legacy proxies. There will only be bug fixes, not new features. This is definitely not a bug fix but rather a new feature, and there won't be a new minor version to release it for.

simPod:
Yup I knew the point is to get rid of common. But I considered this a bug fix as currently it generates invalid proxy method for certain type of identifier getters. So all consumers of Doctrine 2 using same strategy of working with VOs could benefit from this. I saw no harm in fixing it as ultimately it will be all discarded with new major releases of dependent doctrine packages anyway.

lib/Doctrine/Common/Proxy/ProxyGenerator.php
```diff
@@ -20,8 +22,10 @@ class ProxyGenerator
     /**
      * Used to match very simple id methods that don't need
      * to be decorated since the identifier is known.
+     *
+     * @see https://regex101.com/r/u72479/2/
      */
-    const PATTERN_MATCH_ID_METHOD = '((public\s+)?(function\s+%s\s*\(\)\s*)\s*(?::\s*\??\s*\\\\?[a-z_\x7f-\xff][\w\x7f-\xff]*(?:\\\\[a-z_\x7f-\xff][\w\x7f-\xff]*)*\s*)?{\s*return\s*\$this->%s;\s*})i';
+    const PATTERN_MATCH_ID_METHOD = '((public\s+)?(function\s+%s\s*\(\)\s*)\s*(?::\s*\??\s*((?:\\\\?[\w\x7f-\xff][\w\x7f-\xff]*)+)\s*)?{\s*return\s*(\$this->%2$s|new \g{4}\(\$this->%2$s\)|\g{4}::\w+\(\$this->%2$s\));\s*})';
```
greg0ire: This pattern is a bit long, how about we use the `x` modifier?

```suggestion
        const PATTERN_MATCH_ID_METHOD = '(
        (public\s+)?
        (function\s+%s\s*\(\)\s*)\s*
        (?::\s*\??\s*((?:\\\\?[\w\x7f-\xff][\w\x7f-\xff]*)+)\s*)?
        {\s*
            return\s*(\$this->%2$s|new \g{4}\(\$this->%2$s\)|\g{4}::\w+\(\$this->%2$s\));
        \s*}
    )';
```

lib/Doctrine/Common/Proxy/ProxyGenerator.php
```diff
@@ -861,7 +865,7 @@ private function isShortIdentifierGetter($method, ClassMetadata $class)
 
             $pattern = sprintf(self::PATTERN_MATCH_ID_METHOD, $method->getName(), $identifier);
 
-            if (preg_match($pattern, $code)) {
+            if (preg_match(sprintf('~%s~', $pattern), $code)) {
```
greg0ire:
```suggestion
            if (preg_match(sprintf('~%s~x', $pattern), $code)) {
```

greg0ire:
Ah I should have read the comments before reviewing :sweat_smile:

simPod:
@alcaeus do you have any proposition how to move this forward? I don't mean this PR specifically but this feature in general. I'd rather not depend on a fork for a long time.

Like what steps should I take to be able to use doctrine dbal+orm and allow it to lazy load using my VO ids? Thanks!

alcaeus:
> @alcaeus do you have any proposition how to move this forward? I don't mean this PR specifically but this feature in general. I'd rather not depend on a fork for a long time.

Sorry for not replying sooner, but I must have missed this. I'd like to defer this question to @greg0ire or anybody else in @doctrine/doctrinecore as I no longer have the time or energy to maintain this package.

simPod:
Just for the record, using this in prod for almost a year now. Works like a charm.

greg0ire (dismissed):
Sorry, this got out of my radar. Are there docs that need to be updated? Or created?

simPod:
@greg0ire I don't think any doc update is needed. This just fixes parsing of id getters contents.

lib/Doctrine/Common/Proxy/ProxyGenerator.php
```diff
@@ -20,8 +22,19 @@ class ProxyGenerator
     /**
      * Used to match very simple id methods that don't need
      * to be decorated since the identifier is known.
+     *
+     * @see https://regex101.com/r/u72479/2/
      */
-    const PATTERN_MATCH_ID_METHOD = '((public\s+)?(function\s+%s\s*\(\)\s*)\s*(?::\s*\??\s*\\\\?[a-z_\x7f-\xff][\w\x7f-\xff]*(?:\\\\[a-z_\x7f-\xff][\w\x7f-\xff]*)*\s*)?{\s*return\s*\$this->%s;\s*})i';
+    const PATTERN_MATCH_ID_METHOD = <<<REGEXP
```
greg0ire, line 28:
```suggestion
    const PATTERN_MATCH_ID_METHOD = <<<'REGEXP'
```

lib/Doctrine/Common/Proxy/ProxyGenerator.php
```diff
@@ -20,8 +22,19 @@ class ProxyGenerator
...
+     * @see https://regex101.com/r/u72479/2/
      */
-    const PATTERN_MATCH_ID_METHOD = '((public\s+)?(function\s+%s\s*\(\)\s*)\s*(?::\s*\??\s*\\\\?[a-z_\x7f-\xff][\w\x7f-\xff]*(?:\\\\[a-z_\x7f-\xff][\w\x7f-\xff]*)*\s*)?{\s*return\s*\$this->%s;\s*})i';
+    const PATTERN_MATCH_ID_METHOD = <<<'REGEXP'
+~(
+(public\s+)?
+(function\s+%s\s*\(\)\s*)\s*
+(?::\s*\??\s*((?:\\\\?[\w\\x7f-\\xff][\w\\x7f-\\xff]*)+)\s*)?
+{\s*
+return\s*(\\\$this->%2\$s|new\s+\g{4}\(\\\$this->%2\$s\)|\g{4}::\w+\(\\\$this->%2\$s\));
```
greg0ire, line 34: Ah wait, that `$s` needs to be interpreted actually, doesn't it? I'm not sure because before you changed this, it was using single quotes… or is the CI failure about something else entirely?

simPod: yes, tests say so :)

greg0ire (approved):

SenseException:
> TBH, I don't see this getting merged: the entirety of what's left in doctrine/common is deprecated, especially these legacy proxies. There will only be bug fixes, not new features. This is definitely not a bug fix but rather a new feature, and there won't be a new minor version to release it for.

Did this PR become a bugfix so it can be merged? I'm confused.

ostrolucky (changes requested):
I'm sorry but I'm on board with @alcaeus. This is not fixing anything, it's improvement that you call fix. Removing undesired initialization is not a fix, this proxy does not care about non-simple identifier getters by design and by supporting it we open can of worms.

greg0ire:
Thanks for your input everyone, let's close.

simPod:
Can you brief me a bit please. I'm not sure I have the information I need to make a decision on a direction I should take with software I maintain. 

What is the plan here with proxies? Will they be part of the doctrine's other package or will they disappear or? Currently there's this issue but the fix here will not be merged. So proxies are not feasible when one uses more sofisticated VO handling in entities. I'll maintain this fork for now but what is the proper way to fix this issue if this is not going to be merged. What is the proper place to fix this? Thank you.

ostrolucky:
ORM 3 uses https://github.com/Ocramius/ProxyManager, all improvements go there. Until it's released, you will have to use your fork if you need this.

simPod:
OK I'll look into that, thanks for info.

simPod:
Hm but still, ORM 3 will not get released any time soon. New features are being merged https://github.com/doctrine/common/pull/910. TBH I believe this "fix" could have been shipped too...

alcaeus:
Yes, ORM 3 will be out a while. When making the transition to ProxyManager in MongoDB ODM, I realised that introducing this change in a major release causes unnecessary friction and makes it more difficult to update. I believe that the new proxy mechanism should be available in ORM 2.x as well - it's a matter of taking the work and porting it into 2.x to run side-by-side with the legacy proxies. ProxyManager does not hook into method call like the Doctrine proxy library does, but rather initialises the proxy object when uninitialised properties are accessed. This means that methods that don't need to know about the object state can be called without requiring proxy initialisation.

If you are up for the challenge, please let me know, I'd be happy to assist with knowledge I acquired while porting the ORM 3 solution into MongoDB ODM.

Edit: the change in #910 was required to allow people to work with PHP 8 and was thus considered a bugfix. This change would make a significant number of assumptions about the internal workings of user-defined entities which can't be guaranteed, so there may be severe side effects from skipping initialisation based on method names.

simPod:
So are you saying there's new ProxyManager  in ORM 3 that we wish to implement into ORM 2 and you've already implemented it into MongoDB ODM?

If so, can you mention relevant PRs or commits? Thanks!

alcaeus:
For ODM, the main PR was https://github.com/doctrine/mongodb-odm/pull/1875. Note that this already drops the doctrine proxies, which we shouldn't do for ORM. I'm also working on some improvements to doctrine/persistence to make entity class name resolution from proxies more straightforward. If you're interested in tackling this, please reach out to me and I'll assist you any way I can.

simPod:
@alcaeus I checked ORM repo and there are already people looking into dropping doctrine proxies so I'll just use the fork for now.

PR #880, opened 2019-12-18
Address deprecations from doctrine/persistence

greg0ire:
The `Common` portion of the namespace has been removed following a split
of `doctrine/common` into several packages.


greg0ire:
I feel stupid, I updated the lock file but not the json manifest :man_facepalming: I just pushed with that change.

alcaeus:
Note: I'd like to hold off on merging this as we've seen build failures downstream after merging down changes in doctrine/persistence. I'd like to avoid this over the holidays to give people time to unwind.

PR #884, opened 2020-02-28
ProxyGenerator: remove confusing name convention for simple ID getters

develancer:
Currently, ProxyGenerator supports generating simple ID getters (without entity hydration) in proxy class only if method name exactly matches the field name—for example: `$locationIdentifier` with `getLocationIdentifier`. This PR enables this functionality in a broader range of scenarios, e.g.

```
class Entity
{
    protected $location_id;

    public function getLocationIdentifier()
    {
        return $this->location_id;
    }
}
```

This is a very small change in the code, but I believe this would be a significant improvement. Moreover, I would consider it a **bugfix**, as this required naming convention hasn't been documented anywhere.


ostrolucky:
Same story as https://github.com/doctrine/common/pull/878, sorry closing

PR #890, opened 2020-05-14
Switch to persistence 2

greg0ire:
This is a BC-break, because packages consuming doctrine/common expect
persistence types to be defined in the Doctrine\Common namespace, and
they disappear with this change.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -26,7 +26,7 @@
         "doctrine/annotations": "^1.0",
         "doctrine/event-manager": "^1.0",
         "doctrine/reflection": "^1.0",
-        "doctrine/persistence": "^1.3.3"
+        "doctrine/persistence": "^2.0"
     },
     "require-dev": {
         "phpstan/phpstan": "^0.11",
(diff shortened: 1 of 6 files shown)


composer.json
```diff
@@ -26,7 +26,7 @@
         "doctrine/annotations": "^1.0",
         "doctrine/event-manager": "^1.0",
         "doctrine/reflection": "^1.0",
-        "doctrine/persistence": "^1.3.3"
+        "doctrine/persistence": "^2.0"
```
greg0ire, line 29: I'm not using `^1.3.3 ||` here, because removing would mean another BC-break, so another major release. The plan is to have orm use common `2 || 3`, not `3`.

alcaeus: On the other hand, there might either be more releases to allow for major versions of other packages, or no more major releases because we can phase this out completely. Might as well allow `^1.3.3 || ^2.0` for the time being. Other packages can always limit this to `^2.0` only.

greg0ire: I did that mostly because of your comment on the API. If we have that `||`, we have no clear contract over types in the `Common` namespace: they may or may not be there.

alcaeus: Got it. Either way works for me.

PR #893, opened 2020-05-18
Drop packages available stand-alone

alcaeus:
With the migration to persistence 2.0 requiring a new major release, it would be a good time to drop all packages that we've extracted into standalone packages. This keeps necessary functionality (e.g. proxy logic) around and supported, while at the same time removing the need for future major releases when another package (e.g. Annotations, Collections, etc.) release a new major version.

Note that we have for a while suggested that people don't rely on doctrine/common for functionality like annotations, inflector, etc., so this BC break in a major release should not be a big burden on the ecosystem.

Regardless, packages still depending on doctrine/common are encouraged to depend on `"doctrine/common": "^2.13 || ^3.0"` as well as the package for the functionality they need while we make this transition.


greg0ire:
You can get a stack trace for the deprecation by using `SYMFONY_DEPRECATIONS_HELPER=/toString/`

stof:
I'm totally in favor of this change.

alcaeus:
The deprecations were caused due to using lowest dependencies in the lock file, which means testing PHPUnit 7.0 on PHP 7.4. I've removed the lock file altogether and added a build stage to test against lowest deps. Everywhere else we now test highest deps.

.travis.yml
```diff
@@ -22,6 +22,11 @@ script: ./vendor/bin/phpunit
 
 jobs:
   include:
+    - stage: Test
+      php: 7.1
+      install:
+        - travis_retry composer -n --prefer-source --prefer-lowest update
```
stof, line 28: why using `--prefer-source` (while persisting the composer cache in the Travis cache at the same time) ? Using `--prefer-dist` will benefit from the persisted cache.

SenseException: BTW: Using `--no-interaction` is more descriptive than just `-n` and makes the travis-file easier to read.

alcaeus: Good point. I’ll update this 👍

.travis.yml
```diff
@@ -16,12 +16,17 @@ before_install:
   - mv ~/.phpenv/versions/$(phpenv version-name)/etc/conf.d/xdebug.ini{,.disabled} || echo "xdebug not available"
 
 install:
-  - travis_retry composer -n --prefer-source install
+  - travis_retry composer update --no-interaction --prefer-dist  ${COMPOSER_FLAGS}
```
greg0ire, line 19: Shouldn't this be done for the prefer-lowest step and only that one? I'm all for not versioning `composer.lock`, but this kind of change should happen on all branches, otherwise it's confusing for contributors.

alcaeus: Without a committed lock file, `composer update` and `composer install` are the same. The reason why we're using `update` here is so we can pass the `--prefer-lowest` option below without duplicating the entire command. `install` (understandably) doesn't support `--prefer-lowest`.

alcaeus: With this change applied, we're testing latest dependencies in all stages and include one stage that specifically tests against the lowest set of dependencies.

greg0ire: It's great, but it's only on `master`, that's what I'm complaining about here.

alcaeus: I made a bit of a mess, but the change is now in 2.13.x and master. Wanted to put it up for review, then realised I don't own a fork of this.

SenseException: I just wondered why 2.13.x looks similar to master. :-D

PR #894, opened 2020-05-25
Allow PHP 8

greg0ire:


greg0ire:
The trick does not seem to work… let's try `composer update nothing`

greg0ire:
Ah that would be because it's one of these version-composer.lock, use-composer-install packages.

greg0ire:
The failure is caused by a deprecation, ok to fix later I think.

PR #895, opened 2020-05-29
Update deprecations

alcaeus:
* Undeprecate proxy logic: since this is still in active use in ORM 2, there's no point bothering folks with deprecations. We can always revisit these deprecations once the logic has been dropped from ORM.
* Update deprecation message for ClassLoader: we missed dropping this in 3.0, so I'm updating the deprecation message to account for this.


greg0ire:
Thanks @alcaeus !

PR #908, opened 2020-11-19
Proxies inherit return type hints from parent's __sleep and __wakeup methods

VBelozyorov:
If entity's `__sleep` or `__wakeup` methods have return type hints, those hints omitted in proxy classes generated for this entity.
Which produces the error about incompatible method signature.

@alcaeus (?) confirmed that this probably should be changed, short chat about it may be found [here](https://symfony-devs.slack.com/archives/C3FQPE6LE/p1605531655079600)

### DIFF
--- a/tests/Doctrine/Tests/Common/Proxy/ProxyMagicMethodsTest.php
+++ b/tests/Doctrine/Tests/Common/Proxy/ProxyMagicMethodsTest.php
@@ -9,6 +9,8 @@ use InvalidArgumentException;
 use PHPUnit\Framework\MockObject\MockObject;
 use PHPUnit\Framework\TestCase;
 use ReflectionClass;
+use ReflectionNamedType;
+use ReflectionType;
 use function class_exists;
 use function in_array;
 use function serialize;
@@ -295,11 +297,13 @@ class ProxyMagicMethodsTest extends TestCase
         $proxyReflection = new ReflectionClass($proxy);
         self::assertTrue($proxyReflection->hasMethod($sleepMethod), $sleepMethod . 'method doesn\'t exist');
 
-        $methodReflection = $proxyReflection->getMethod($sleepMethod);
+        $methodReflection     = $proxyReflection->getMethod($sleepMethod);
+        $returnTypeReflection = $methodReflection->getReturnType();
+        self::assertInstanceOf(ReflectionType::class, $returnTypeReflection, 'Got unexpected return type reflection');
         self::assertSame(
-            $methodReflection->getReturnType()->getName(),
+            $returnTypeReflection instanceof ReflectionNamedType ? $returnTypeReflection->getName() : (string) $returnTypeReflection,
             'array',
-            $sleepMethod . ' method has lost its return type hint'
+            $sleepMethod . ' method has lost return type hint'
         );
     }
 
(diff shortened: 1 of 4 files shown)


VBelozyorov:
Actually `./vendor/bin/phpunit` runs just fine with no warnings or errors related to methods in this PR, so what should I do with failed PHPStan check?

tests/Doctrine/Tests/Common/Proxy/ProxyMagicMethodsTest.php
```diff
@@ -285,6 +285,24 @@ public function testInheritedMagicSleep()
...
+        $proxy          = new $proxyClassName();
+
+        $sleepMethod = '__sleep';
+
+        $proxyReflection = new ReflectionClass($proxy);
+        self::assertTrue($proxyReflection->hasMethod($sleepMethod), $sleepMethod . 'method doesn\'t exist');
+
+        $methodReflection = $proxyReflection->getMethod($sleepMethod);
+        self::assertSame(
+            $methodReflection->getReturnType()->getName(),
```
greg0ire: What guarantees that this will return a `ReflectionNamedType`, not just a `ReflectionType`?

alcaeus: `getReturnType` will return a `ReflectionNamedType` in the supported PHP versions, unless we're on PHP 8 and the method has a union return type, in which case this is a `ReflectionUnionType`. So there should be a conditional to support both, as `ReflectionUnionType` doesn't have a `getName` method.

VBelozyorov: Should we check something in case of `ReflectionUnionType`?
In context of `__sleep` and `__wakeup` there not so much variability, I suppose.

VBelozyorov: [Here](https://wiki.php.net/rfc/union_types_v2#reflection) I see `__toString` method but it confuses me that it is deprecated in parent `ReflectionType` class.

alcaeus: Yes, the method is deprecated. For `ReflectionNamedType` (which is always returned for `getReturnType` starting with PHP 7.1.0), you should use the `getName()` method. Not sure what the alternative is for `ReflectionUnionType`.

tests/Doctrine/Tests/Common/Proxy/ProxyMagicMethodsTest.php
```diff
@@ -285,6 +285,24 @@ public function testInheritedMagicSleep()
...
+            'array',
+            $sleepMethod . ' method has lost its return type hint'
+        );
```
VBelozyorov, line 303: Is deprecated method better than nothing? :wink: 
```suggestion
        $returnTypeReflection = $methodReflection->getReturnType();
        self::assertInstanceOf(\ReflectionType::class, $returnTypeReflection, 'Got unexpected return type reflection');
        if ($returnTypeReflection instanceof \ReflectionNamedType) {
            self::assertSame(
                $methodReflection->getReturnType()->getName(),
                'array',
                $sleepMethod . ' method has lost return type hint'
            );
        } else {
            self::assertSame(
                (string) $returnTypeReflection,
                'array',
                $sleepMethod . ' method has lost return type hint'
            );
        }
```

alcaeus (approved):

greg0ire (approved):

greg0ire:
Thanks @VBelozyorov !

VBelozyorov:
I am glad to help. Is there any chance to merge this into 2.13.x? Can I do something for it? 
Because initially I ran into this problem in a project where `doctrine/common 2.13.3` is used.

greg0ire:
We'd rather not, sorry… what's stopping you from upgrading to 3.0? If there is a major blocker we might reconsider.

PR #910, opened 2020-11-26
Add support for union types and mixed type in ProxyGenerator

beberlei:
Adds support for generating proxies for classes using the following new PHP 8 syntax constructs:

- Union Types
- Mixed type

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -24,7 +24,7 @@
     "require-dev": {
         "phpstan/phpstan": "^0.12",
         "phpstan/phpstan-phpunit": "^0.12",
-        "phpunit/phpunit": "^7.5.20 || ^8.0 || ^9.0",
+        "phpunit/phpunit": "^7.5.20 || ^8.5 || ^9.0",
         "doctrine/coding-standard": "^6.0 || ^8.0",
         "squizlabs/php_codesniffer": "^3.0",
         "symfony/phpunit-bridge": "^4.0.5"
--- a/lib/Doctrine/Common/Proxy/ProxyGenerator.php
+++ b/lib/Doctrine/Common/Proxy/ProxyGenerator.php
@@ -10,6 +10,8 @@ use ReflectionMethod;
 use ReflectionNamedType;
 use ReflectionParameter;
 use ReflectionProperty;
+use ReflectionType;
+use ReflectionUnionType;
 
 use function array_combine;
 use function array_diff;
@@ -1112,6 +1114,15 @@ EOT;
         ReflectionMethod $method,
         ?ReflectionParameter $parameter = null
     ) {
+        if ($type instanceof ReflectionUnionType) {
+            return implode('|', array_map(
+                function (ReflectionType $unionedType) use ($method, $parameter) {
+                    return $this->formatType($unionedType, $method, $parameter);
+                },
+                $type->getTypes()
+            ));
+        }
+
         $name      = $type->getName();
         $nameLower = strtolower($name);
 
@@ -1145,6 +1156,7 @@ EOT;
         if (
             $type->allowsNull()
             && ($parameter === null || ! $parameter->isDefaultValueAvailable() || $parameter->getDefaultValue() !== null)
+            && $name !== 'mixed'
         ) {
             $name = '?' . $name;
         }
--- a/lib/Doctrine/Common/Proxy/ProxyGenerator.php
+++ b/lib/Doctrine/Common/Proxy/ProxyGenerator.php
@@ -7,7 +7,6 @@ use Doctrine\Common\Proxy\Exception\UnexpectedValueException;
 use Doctrine\Common\Util\ClassUtils;
 use Doctrine\Persistence\Mapping\ClassMetadata;
 use ReflectionMethod;
-use ReflectionNamedType;
 use ReflectionParameter;
 use ReflectionProperty;
 use ReflectionType;
@@ -1110,7 +1109,7 @@ EOT;
      * @return string
      */
     private function formatType(
-        ReflectionNamedType $type,
+        ReflectionType $type,
         ReflectionMethod $method,
         ?ReflectionParameter $parameter = null
     ) {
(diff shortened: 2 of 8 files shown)


composer.json
```diff
@@ -24,8 +24,8 @@
     "require-dev": {
         "phpstan/phpstan": "^0.11",
         "phpstan/phpstan-phpunit": "^0.11",
-        "phpunit/phpunit": "^7.0",
-        "doctrine/coding-standard": "^1.0",
+        "phpunit/phpunit": "^7.0|^8.5|^9.4",
+        "doctrine/coding-standard": "^6.0|^8.0",
```
beberlei: i will make a seperate PR that updates coding standard first and autofixes everything.

lib/Doctrine/Common/Proxy/ProxyGenerator.php
```diff
@@ -1141,6 +1150,7 @@ private function formatType(
 
         if ($type->allowsNull()
             && (null === $parameter || ! $parameter->isDefaultValueAvailable() || null !== $parameter->getDefaultValue())
+            && $name !== 'mixed'
```
beberlei, line 1161: mixed returns true in `allowsNull` but if you use it with `?mixed` it throws a fatal error, so this adds a special case.

tests/Doctrine/Tests/Common/Proxy/ProxyGeneratorTest.php
```diff
@@ -396,6 +398,48 @@ public function testFinalClassThrowsException()
         $proxyGenerator->generateProxyClass($this->createClassMetadata(FinalClass::class, []));
     }
 
+    public function testPhp8UnionTypes()
+    {
+        if (PHP_VERSION_ID < 80000) {
+            $this->markTestSkipped('Requires PHP 8.0');
+        }
+
```
greg0ire: I never used it myself, but apparently there is an annotation for this:

```suggestion
    /**
     * @requires PHP >= 8.0.0
     */
    public function testPhp8UnionTypes()
    {
```

tests/Doctrine/Tests/Common/Proxy/ProxyGeneratorTest.php
```diff
@@ -396,6 +398,48 @@ public function testFinalClassThrowsException()
...
+            file_get_contents(__DIR__ . '/generated/__CG__DoctrineTestsCommonProxyPhp8UnionTypes.php')
+        );
+    }
+
+    public function testPhp8MixedType()
+    {
+        if (PHP_VERSION_ID < 80000) {
+            $this->markTestSkipped('Requires PHP 8.0');
+        }
+
```
greg0ire:
```suggestion
    /**
     * @requires PHP >= 8.0.0
     */
    public function testPhp8MixedType()
    {
```

tests/Doctrine/Tests/Common/Proxy/ProxyMagicMethodsTest.php
```diff
@@ -367,6 +367,10 @@ public function testInheritedMagicIssetWithBoolean()
 
     public function testInheritedMagicIssetWithInteger()
     {
+        if (PHP_VERSION_ID >= 80000) {
+            $this->markTestSkipped('isset returning a non boolean is not supported on PHP 8.0+ anymore.');
+        }
```
greg0ire: Maybe that one should stay like that because of the helpful message :thinking:

lib/Doctrine/Common/Proxy/ProxyGenerator.php
```diff
@@ -1108,10 +1109,19 @@ private function shouldProxiedMethodReturn(ReflectionMethod $method)
...
+        if ($type instanceof ReflectionUnionType) {
+            return implode('|', array_map(
+                function (ReflectionType $unionedType) use ($method, $parameter) {
+                    return $this->formatType($unionedType, $method, $parameter);
+                },
+                $type->getTypes()
+            ));
+        }
+
         $name      = $type->getName();
```
greg0ire, line 1128: Maybe add an `assert` before? If it's not a `ReflectionUnionType`, then it must be a `ReflectionNamedType`, right? (although phpstan can't know that for sure).

Another solution might be to use `@param ReflectionUnionType|ReflectionNamedType $type`, but it might cause other issues, I'm not sure.

lib/Doctrine/Common/Proxy/ProxyGenerator.php
greg0ire: Is that a bug in PHP?

tests/Doctrine/Tests/Common/Proxy/Php8UnionTypes.php
```diff
@@ -0,0 +1,14 @@
+<?php
+
```
greg0ire, line 2: We might want to run Static Analysis with PHP 8.0 from now on.

lib/Doctrine/Common/Proxy/ProxyGenerator.php
beberlei: No why? allowsNul() is right about true, maybe `?mixed` would not have been a big problem, but its right mixed already contains null, so its redundant.

greg0ire: Oh right, indeed.

ondrejmirtes:
BTW if you need any help with PHPStan, don't hesitate to reach me directly instead of speculating over less than ideal solutions...

greg0ire:
Will ping you directly from now on ^^

lib/Doctrine/Common/Proxy/ProxyGenerator.php
```diff
@@ -1108,10 +1110,21 @@ private function shouldProxiedMethodReturn(ReflectionMethod $method)
...
+        ReflectionType $type,
         ReflectionMethod $method,
         ?ReflectionParameter $parameter = null
     ) {
+        if ($type instanceof ReflectionUnionType) {
+            return implode('|', array_map(
+                function (ReflectionType $unionedType) use ($method, $parameter) {
+                    return $this->formatType($unionedType, $method, $parameter);
+                },
+                $type->getTypes()
```
greg0ire, line 1122: CodeCov says the lines are not covered, and it is correct: there is no PHP 8 job yet. I'll try to add one in a different PR, as well as the other minor improvements I added that pollute this PR.

greg0ire: Now I rebased, it no longer complains :)

greg0ire: I rebased, there no longer is a comment :+1:

greg0ire (approved):

PR #911, opened 2020-11-26
Update doctrine/coding-standard to 6.0|8.0

beberlei:
This upgrades the coding-standard and fixes all the new violations to pass it.

Add exclusions for coding styles to all of the test dummy classes needed for proxy as their changes would break tests.


greg0ire:
I was wondering why there were no CI checks, then created https://github.com/doctrine/common/pull/912, which makes the move from 1 to 6. I will rebase your work on it after it gets merged.

greg0ire (approved):

PR #913, opened 2020-11-29
Maintenance work

greg0ire:


greg0ire:
I'm targeting 3.0.x because we don't want bugfixes to break PHP 8 compatibility. This means we have to cherry-pick the CS upgrade otherwise we cannot test with PHP8

SenseException (approved):

PR #918, opened 2021-01-15
Better PHP8 type handling

adambalint-srg:
I've found two type declaration related problems in proxy generator, when using it with PHP8.

First is the usage of static keyword, which should be "copied" to the proxy class, and also can be nullable.
`function setId(int $id): ?static`
This is supported in this pull request.

Second is a fix for union types of PHP8, because in case of nullable union types the '?' sign is nut supported. It should be used in this way:
`float|int|null`
But the proxy generator prepended the '?' to the null part of the union type:
`float|int|?null`
This problem has been fixed with an optional argument in case of recursive call of formatType function.

### DIFF
--- a/lib/Doctrine/Common/Proxy/ProxyGenerator.php
+++ b/lib/Doctrine/Common/Proxy/ProxyGenerator.php
@@ -1112,13 +1112,12 @@ EOT;
     private function formatType(
         ReflectionType $type,
         ReflectionMethod $method,
-        ?ReflectionParameter $parameter = null,
-        bool $isUnionPart = false
+        ?ReflectionParameter $parameter = null
     ) {
         if ($type instanceof ReflectionUnionType) {
             return implode('|', array_map(
                 function (ReflectionType $unionedType) use ($method, $parameter) {
-                    return $this->formatType($unionedType, $method, $parameter, true);
+                    return $this->formatType($unionedType, $method, $parameter);
                 },
                 $type->getTypes()
             ));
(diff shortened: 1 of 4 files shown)


lib/Doctrine/Common/Proxy/ProxyGenerator.php
```diff
@@ -1151,12 +1156,13 @@ function (ReflectionType $unionedType) use ($method, $parameter) {
...
+        if (! $type->isBuiltin() && $name !== 'static') {
             $name = '\\' . $name;
         }
 
         if (
-            $type->allowsNull()
+            ! $isUnionPart
+            && $type->allowsNull()
             && ($parameter === null || ! $parameter->isDefaultValueAvailable() || $parameter->getDefaultValue() !== null)
             && $name !== 'mixed'
```
jellynoone: Thanks for working on this just found the nullable issue myself. 

Would it perhaps be better to avoid creating a new parameter and just check for name null like it is done for mixed?

```php
if (
    $type->allowsNull()
    && ! in_array($name, ['mixed', 'null'], true)
    && ($parameter === null || ! $parameter->isDefaultValueAvailable() || $parameter->getDefaultValue() !== null)
) 
```

adambalint-srg: Thanks for the suggestion, it's more logical :) I've fixed this with a new commit

jellynoone:
Can someone from the Doctrine team review the pull request? @beberlei  @greg0ire 

By now the author, myself and https://github.com/doctrine/orm/issues/8422 have found the issue being solved in this pull request.

jellynoone (approved):

greg0ire (approved):

SenseException (approved):

beberlei:
Does this target the 3.1.x branch for a specific reason? This is a bugfix for me and should target 3.0.x

Edit: Sorry, my mistake Github showed 3.0.3 as last release, but it was actually 3.1.0 already.

PR #920, opened 2021-01-24
Add Psalm

VincentLanglet:


.github/workflows/static-analysis.yml
```diff
@@ -37,3 +37,29 @@ jobs:
...
+        uses: "shivammathur/setup-php@v2"
+        with:
+          coverage: "none"
+          php-version: "${{ matrix.php-version }}"
+
+      - name: "Install dependencies with Composer"
+        uses: "ramsey/composer-install@v1"
+
+      - name: "Run a static analysis with vimeo/psalm"
+        run: "vendor/bin/psalm"
```
greg0ire: We usually use some extra options: https://github.com/doctrine/.github/blob/2c60b2f943cf08df21be3dc733f1b92db8d24beb/workflow-templates/static-analysis.yml#L63

Are those maybe not required anymore?

.github/workflows/static-analysis.yml
```diff
@@ -37,3 +37,29 @@ jobs:
...
+        continue-on-error: true
```
greg0ire: Why would we want the job to pass if this step fails?

VincentLanglet: I don't know, this was like this for phpstan. So I supposed you didn't want it to be a required build
Fixed

.github/workflows/static-analysis.yml
VincentLanglet: Done

greg0ire (approved):

greg0ire:
I did not spot that phpstan was allowed to fail in https://github.com/doctrine/common/pull/910, maybe we could not find another way? An ignore rule would have been better. I'd fix it but I'm on my phone.

VincentLanglet:
> I did not spot that phpstan was allowed to fail in #910, maybe we could not find another way? An ignore rule would have been better. I'd fix it but I'm on my phone.

Phpstan errors are:
```
 ------ --------------------------------------------------------------------
  Line   lib/Doctrine/Common/Proxy/ProxyGenerator.php
 ------ --------------------------------------------------------------------
  1119   Class ReflectionUnionType not found.
         💡 Learn more at https://phpstan.org/user-guide/discovering-symbols
  1124   Call to method getTypes() on an unknown class ReflectionUnionType.
         💡 Learn more at https://phpstan.org/user-guide/discovering-symbols
 ------ --------------------------------------------------------------------

 ------ --------------------------------------------------------------------
  Line   tests/Doctrine/Tests/Common/Proxy/ProxyMagicMethodsTest.php
 ------ --------------------------------------------------------------------
  417    Class Doctrine\Tests\Common\Proxy\MagicIssetClassWithInteger not
         found.
         💡 Learn more at https://phpstan.org/user-guide/discovering-symbols
 ------ --------------------------------------------------------------------
```

I think it's better to fix it in another PR

greg0ire:
Me too FTR

SenseException (approved):
Yes, it can be handled in another PR.

greg0ire:
Thanks @VincentLanglet !

VincentLanglet:
Hi @greg0ire does a 3.1.2 release would be possible ? :)

greg0ire:
Hi! Sure! https://github.com/doctrine/common/releases/tag/3.1.2

VincentLanglet:
Thanks

PR #921, opened 2021-01-25
Fix ProxyGenerator magic __set with void return type

### LINKED ISSUE #917: ProxyGenerator __set() with return type hint void always returns parent
As the title suggests, I noticed that the generated proxy class is breaking my entity when it contains the magic `__set()` method.

As of now the `ProxyGenerator` will always generate a `return parent::__set(...)` call in the proxy class. Example:

(Original entity)
```php
    /**
     * @inheritDoc
     */
    public function __get(string $name)
    {
        return $this->store[$name];
    }

    /**
     * @inheritDoc
     */
    public function __set(string $name, $value): void
    {
        $this->store[$name] = $value;
    }
```

(Proxy entity)
```php
    /**
     * {@inheritDoc}
     * @param string $name
     */
    public function __get(string $name)
    {
        $this->__initializer__ && $this->__initializer__->__invoke($this, '__get', [$name]);
        return parent::__get($name);
    }

    /**
     * {@inheritDoc}
     * @param string $name
     * @param mixed  $value
     */
    public function __set(string $name, $value): void
    {
        $this->__initializer__ && $this->__initializer__->__invoke($this, '__set', [$name, $value]);

        return parent::__set($name, $value); // <-- Does not work with 'void' return type.
    }
```

Upon invocation this code will produce the following error message:

> Compile Error: A void function must not return a value

(Original entity, temporary workaround)
```php
    /**
     * @inheritDoc
     */
    public function __get(string $name)
    {
        return $this->store[$name];
    }

    /**
     * @inheritDoc
     */
    public function __set(string $name, $value)/*: void*/
    {
        $this->store[$name] = $value;
    }
```

I'm using `doctrine/common` version `3.0.2`, but I've also checked back on `master`. PHP `8.0.0`, Xdebug `3.0.1`. I've already tracked down the cause of this problem and a bunch of closed issues related to the generated magic methods, listed below.

* https://github.com/doctrine/orm/issues/6221
* https://github.com/doctrine/common/blob/3.1.x/lib/Doctrine/Common/Proxy/ProxyGenerator.php#L593
* https://github.com/doctrine/common/issues/868
* https://github.com/doctrine/common/pull/869/files

As shown by the generated code above and the changes made in the linked PR, the `__get()` magic methods works just fine with return types and also has a check in place to avoid this issue:

```php
            if ($returnTypeHint === ': void') {
                $magicGet .= "\n            return;";
            } else {
                $magicGet .= "\n            return \$this->$name;";
            }
```

I think that check is missing for the `__set()` magic method and it's responsible for this wrong behavior. I'd propose a change to make this:

```php
        if ($hasParentSet) {
            $magicSet .= <<<'EOT'
        $this->__initializer__ && $this->__initializer__->__invoke($this, '__set', [$name, $value]);

        return parent::__set($name, $value);
EOT;
        } else {
            $magicSet .= "        \$this->\$name = \$value;";
        }
```

into this:

```php
        if ($hasParentSet) {
            $magicSet .= "        \$this->__initializer__ && \$this->__initializer__->__invoke(\$this, '__set', [\$name, \$value]);";

            if ($returnTypeHint === ': void') {
                $magicSet .= <<<'EOT'

        parent::__set($name, $value);
        return;
EOT;
            } else {
                $magicSet .= <<<'EOT'

        return parent::__set($name, $value);
EOT;
            }
        } else {
            $magicSet .= "        \$this->\$name = \$value;";
        }
```

Any thoughts on this? Maybe I'm missing something here.

GameplayJDK:
See #917 for details.

### DIFF
--- a/lib/Doctrine/Common/Proxy/ProxyGenerator.php
+++ b/lib/Doctrine/Common/Proxy/ProxyGenerator.php
@@ -564,7 +564,7 @@ EOT
             return '';
         }
 
-        $magicSet   = <<<EOT
+        $magicSet = <<<EOT
     /**
      * $inheritDoc
      * @param string \$name
--- a/lib/Doctrine/Common/Proxy/ProxyGenerator.php
+++ b/lib/Doctrine/Common/Proxy/ProxyGenerator.php
@@ -545,12 +545,17 @@ EOT
     private function generateMagicSet(ClassMetadata $class)
     {
         $lazyPublicProperties = $this->getLazyLoadedPublicPropertiesNames($class);
-        $hasParentSet         = $class->getReflectionClass()->hasMethod('__set');
+        $reflectionClass      = $class->getReflectionClass();
+        $hasParentSet         = false;
+        $inheritDoc           = '';
         $parametersString     = '$name, $value';
         $returnTypeHint       = null;
 
-        if ($hasParentSet) {
-            $methodReflection = $class->getReflectionClass()->getMethod('__set');
+        if ($reflectionClass->hasMethod('__set')) {
+            $hasParentSet     = true;
+            $inheritDoc       = '{@inheritDoc}';
+            $methodReflection = $reflectionClass->getMethod('__set');
+
             $parametersString = $this->buildParametersString($methodReflection->getParameters(), ['name', 'value']);
             $returnTypeHint   = $this->getMethodReturnType($methodReflection);
         }
@@ -559,18 +564,16 @@ EOT
             return '';
         }
 
-        $inheritDoc = $hasParentSet ? '{@inheritDoc}' : '';
-        $magicSet   = sprintf(<<<'EOT'
+        $magicSet   = <<<EOT
     /**
-     * %s
-     * @param string $name
-     * @param mixed  $value
+     * $inheritDoc
+     * @param string \$name
+     * @param mixed  \$value
      */
-    public function __set(%s)%s
+    public function __set($parametersString)$returnTypeHint
     {
 
-EOT
-            , $inheritDoc, $parametersString, $returnTypeHint);
+EOT;
 
         if (! empty($lazyPublicProperties)) {
             $magicSet .= <<<'EOT'
@@ -589,9 +592,20 @@ EOT;
         if ($hasParentSet) {
             $magicSet .= <<<'EOT'
         $this->__initializer__ && $this->__initializer__->__invoke($this, '__set', [$name, $value]);
+EOT;
+
+            if ($returnTypeHint === ': void') {
+                $magicSet .= <<<'EOT'
+
+        parent::__set($name, $value);
+        return;
+EOT;
+            } else {
+                $magicSet .= <<<'EOT'
 
         return parent::__set($name, $value);
 EOT;
+            }
         } else {
             $magicSet .= '        $this->$name = $value;';
         }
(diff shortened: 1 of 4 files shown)


SenseException:
Please create a test that covers the bug behavior.

tests/Doctrine/Tests/Common/Proxy/MagicSetClassWithScalarTypeAndRenamedParameters.php
```diff
@@ -24,7 +24,7 @@ class MagicSetClassWithScalarTypeAndRenamedParameters
      *
      * @throws BadMethodCallException
      */
-    public function __set($n, $val)
```
SenseException: This class is part of a different test and it should keep `__set($n, $val)` to cover cases where no types are used. Your case is already covered with `MagicSetClassWithScalarTypeAndVoidReturnType`.

SenseException (changes requested):

tests/Doctrine/Tests/Common/Proxy/MagicSetClassWithScalarTypeAndRenamedParameters.php
GameplayJDK: I'll undo that change, but the `__set()` there should definitely have a scalar type in it's signature, otherwise the name of the test class is misleading. As I understood it, the `MagicSetClass` class does not have any type in its signature, while the `MagicSetClassWithScalarType` class does have a type in its signature, hence the name. So the `MagicSetClassWithScalarTypeAndRenamedParameters` should either be renamed to `MagicSetClassWithRenamedParameters` or have the type added, to make this consistent. But that could also be proposed in yet another PR, as it does not impact the changes introduced with this one.

SenseException: I don't know about the intention of the naming, so let's keep those names for now, while keeping the original state of this `__set()`. Your new test will cover the type.

GameplayJDK: Agreed. I'm going to revert this change now and hand in the PR for a re-review :)

lib/Doctrine/Common/Proxy/ProxyGenerator.php
```diff
@@ -559,18 +564,16 @@ private function generateMagicSet(ClassMetadata $class)
...
         }
 
-        $inheritDoc = $hasParentSet ? '{@inheritDoc}' : '';
-        $magicSet   = sprintf(<<<'EOT'
+        $magicSet = <<<EOT
     /**
-     * %s
-     * @param string $name
-     * @param mixed  $value
+     * $inheritDoc
```
SenseException, line 569: I suppose you meant 
```suggestion
     * {@inheritDoc}
```

SenseException (changes requested):

lib/Doctrine/Common/Proxy/ProxyGenerator.php
GameplayJDK: As with the `generateMagicGet()` implementation, the `$inheritDoc` is set to `{@inheritDoc}` if `$reflectionClass->hasMethod('__set')` evaluates to `true`, see line 556.

SenseException (approved):
I'd prefer a second Doctrine reviewer who's taking a look at this. From my pov this looks good.

malarzm (approved):

malarzm:
Thanks @GameplayJDK!

PR #925, opened 2021-02-22
added prepend parameter for the autoload register

Aazhar:
when using some other auto-loaders, we would like to prepend doctrine autoloader to avoid errors


SenseException:
It's an internal class so I assume this change is okay. A new test that checks a prepended autoloader would be helpful.

SenseException (changes requested):
It's an internal class so I assume this change is okay. A new test that checks a prepended autoloader would be helpful.

PR #926, opened 2021-04-02
Add generics to ClassUtils::getClass

franmomu:
This PR tries to add generics to `ClassUtils::getClass` and `ClassUtils::getRealClass`, but what I'm not really sure is if this is strictly right since calling this method using a proxy instance, it won't return the `class-string` of the proxy but the real class.

Calling it with an instance of: `MyProxy\Subdir\__CG__\RealClass` would return `RealClass`.

### DIFF
--- a/lib/Doctrine/Common/Util/ClassUtils.php
+++ b/lib/Doctrine/Common/Util/ClassUtils.php
@@ -25,8 +25,9 @@ class ClassUtils
      *
      * @return string
      *
-     * @psalm-param class-string $className
-     * @psalm-return class-string
+     * @template T of object
+     * @psalm-param class-string<T> $className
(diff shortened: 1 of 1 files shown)


stof:
I don't think the current generics are right, precisely due to the fact that it returns a different class-string

franmomu:
> I don't think the current generics are right, precisely due to the fact that it returns a different class-string

👍 Makes sense

VincentLanglet:
What about making a generic proxy ? @franmomu 

```
@psalm-template T of object
@psalm-param class-string<Proxy<T>|T>
@psalm-return class-string<T>
```

```
@psalm-template T of object
@psalm-param Proxy<T>|T
@psalm-return class-string<T>
```

PR #928, opened 2021-04-24
Fixed doctrine/persistence dependency

smilesrg:
`doctrine/persistence` dependency has been updated to `3.0@dev` version for `3.0.x` branch.

That problem leads to 

```
 Your lock file does not contain a compatible set of packages. Please run composer update.
 
   Problem 1
     - doctrine/common is locked to version 3.0.2 and an update of this package was not requested.
     - doctrine/common 3.0.2 requires doctrine/persistence ^2.0 -> found doctrine/persistence[3.0.x-dev] but it does not match the constraint.
```

while installing dependency for doctrine/persistence

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -19,7 +19,7 @@
     ],
     "require": {
         "php": "^7.1 || ^8.0",
-        "doctrine/persistence": "^2.0"
+        "doctrine/persistence": "^3.0@dev"
     },
     "require-dev": {
         "phpstan/phpstan": "^0.12",


greg0ire:
> doctrine/persistence dependency has been updated to 3.0@dev version for 3.0.x branch.
> That problem leads to

I'm sorry, what problem? And why would we use a dev version on a stable branch? Also, 3.0.x is no longer maintained.

smilesrg:
>I'm sorry, what problem? 

I described it in the description (I've just updated it). I tried to `composer install` on `doctrine/persistence 3.0.x` but got an errors with dependencies.

>And why would we use a dev version on a stable branch? 

 If remove `@dev` prefix, you can get minimum stability error while doing `composer install` on `doctrine/common`

>Also, 3.0.x is no longer maintained.

Could you please tell which branch is proper to use?

smilesrg:
>Also, 3.0.x is no longer maintained.

There's no information about this
https://www.doctrine-project.org/projects/persistence.html

smilesrg:
Well, dependencies are totally confused. I am working on https://github.com/doctrine/orm/pull/8657  issue https://github.com/doctrine/orm/issues/8460 (Deprecate EntityManager::clear with argument of one entity class and EntityRepository::clear), making changes in ORM 3.x but I also need to make changes in `doctrine/persistence` https://github.com/doctrine/persistence/pull/174, I guess I should make changes in `3.x` branch? But then, if I am using `3.x` branch, error appears during `composer install`. Also, it is so strange, when stable `doctrine/common 3.x` stable branches are using persistence `2.x` stable branch. I really confused with the dependencies. Could you please help me to resolve that?

greg0ire:
You can use the following to trick `doctrine/persistence` into thinking it is in version 2.0 instead of 3:

`COMPOSER_ROOT_VERSION=2.0 composer install --no-interaction --no-progress`

More details here on why I added this dependency: https://github.com/doctrine/persistence/pull/122#discussion_r462586031

> Could you please tell which branch is proper to use?

If it's a fix, 3.1.x. `.doctrine-project.json` is often outdated, so I'll give you the rule: if the last tag is X.Y.Z, then fixes should got to X.Y, and features to X.Y+1 .

PR #938, opened 2021-10-13
ProxyFactory > Add `FILE_NOT_EXISTS_OR_CHANGED` mode

ruudk:
We have a fairly large Symfony application with ±230 entities.

While profiling for performance bottlenecks in development I noticed that proxies were always
regenerated. This was due the fact that we run the application in debug mode and by default, the
DoctrineBundle chooses `AUTOGENERATE_ALWAYS` when run in debug mode.

After changing the value to `AUTOGENERATE_FILE_NOT_EXISTS` I was able to shave of 37ms of a single
query. Great!

But this has a big drawback. Whenever we make changes to the proxied entity, the proxy is never
updated anymore.

To solve this, I think it makes sense to look at the file modification times of both the proxy and
the proxied entity.

When the proxied entity is newer than the proxy, it should regenerate the proxy.

Screenshot 2021-10-13 at 09 29 46@2x

## Todo

- [ ] When this is merged, add support to DoctrineBundle

### DIFF
--- a/lib/Doctrine/Common/Proxy/AbstractProxyFactory.php
+++ b/lib/Doctrine/Common/Proxy/AbstractProxyFactory.php
@@ -10,6 +10,7 @@ use Doctrine\Persistence\Mapping\ClassMetadataFactory;
 
 use function class_exists;
 use function file_exists;
+use function filemtime;
 use function in_array;
 use function interface_exists;
 
@@ -34,7 +35,7 @@ abstract class AbstractProxyFactory
     /**
      * Autogenerate the proxy class when the proxy file does not exist.
      *
-     * This strategy causes a file exists call whenever any proxy is used the
+     * This strategy causes a file_exists() call whenever any proxy is used the
      * first time in a request.
      */
     public const AUTOGENERATE_FILE_NOT_EXISTS = 2;
@@ -47,11 +48,22 @@ abstract class AbstractProxyFactory
      */
     public const AUTOGENERATE_EVAL = 3;
 
+    /**
+     * Autogenerate the proxy class when the proxy file does not exist or
+     * when the proxied file changed.
+     *
+     * This strategy causes a file_exists() call whenever any proxy is used the
+     * first time in a request. When the proxied file is changed, the proxy will
+     * be updated.
+     */
+    public const AUTOGENERATE_FILE_NOT_EXISTS_OR_CHANGED = 4;
+
     private const AUTOGENERATE_MODES = [
         self::AUTOGENERATE_NEVER,
         self::AUTOGENERATE_ALWAYS,
         self::AUTOGENERATE_FILE_NOT_EXISTS,
         self::AUTOGENERATE_EVAL,
+        self::AUTOGENERATE_FILE_NOT_EXISTS_OR_CHANGED,
     ];
 
     /** @var ClassMetadataFactory */
@@ -202,6 +214,14 @@ abstract class AbstractProxyFactory
 
                 case self::AUTOGENERATE_EVAL:
                     $this->proxyGenerator->generateProxyClass($classMetadata, false);
+                    break;
+
+                case self::AUTOGENERATE_FILE_NOT_EXISTS_OR_CHANGED:
+                    if (! file_exists($fileName) || filemtime($fileName) < filemtime($classMetadata->getReflectionClass()->getFileName())) {
+                        $this->proxyGenerator->generateProxyClass($classMetadata, $fileName);
+                    }
+
+                    require $fileName;
                     break;
             }
         }
--- a/tests/Doctrine/Tests/Common/Proxy/AbstractProxyFactoryTest.php
+++ b/tests/Doctrine/Tests/Common/Proxy/AbstractProxyFactoryTest.php
@@ -31,6 +31,7 @@ class AbstractProxyFactoryTest extends DoctrineTestCase
             [1, 1],
             [2, 2],
             [3, 3],
+            [4, 4],
             ['2', 2],
             [true, 1],
             [false, 0],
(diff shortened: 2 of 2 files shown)


greg0ire:
This looks great!

2 remarks though:
- there are no tests. I see there is an existing test class for this class, can you please look into testing what you added?
- there are no docs, but we don't seem to have docs at all, so I can't reasonably require you to write some… if you have some time to spare, please consider writing docs about the proxy system, and if you don't, the phpdoc will be enough :)

lib/Doctrine/Common/Proxy/AbstractProxyFactory.php
```diff
@@ -47,11 +48,22 @@ abstract class AbstractProxyFactory
      */
     public const AUTOGENERATE_EVAL = 3;
 
+    /**
+     * Autogenerate the proxy class when the proxy file does not exist or
+     * when the entity changed.
```
greg0ire: @alcaeus @franmomu is "entity" the correct word here? Or is this class used for the ODM as well? There are 2 other occurrences of "entity" in that project.

lib/Doctrine/Common/Proxy/AbstractProxyFactory.php
```diff
@@ -47,11 +48,22 @@ abstract class AbstractProxyFactory
...
+     *
+     * This strategy causes a file exists call whenever any proxy is used the
```
greg0ire:
```suggestion
     * This strategy causes a file_exists() call whenever any proxy is used the
```

lib/Doctrine/Common/Proxy/AbstractProxyFactory.php
franmomu: 🤔 [since `2.0` ODM is using ProxyManager](https://github.com/doctrine/mongodb-odm/blob/2.2.x/UPGRADE-2.0.md#proxy-objects), so I would say it is not used in the ODM.

But if the PR is about files, I guess I would use `proxied file` or similar.

ruudk: Fixed

ruudk:
Added some tests. Unfortunately, I had to add a sleep to make sure the `filemtime` calls work properly.

tests/Doctrine/Tests/Common/Proxy/AbstractProxyFactoryTest.php
```diff
@@ -215,6 +216,170 @@ public function testMissingPrimaryKeyValue()
...
+            });
+
+        /** @var MockObject&AbstractProxyFactory $proxyFactory */
+        $proxyFactory = $this->getMockForAbstractClass(
+            AbstractProxyFactory::class,
+            [$proxyGenerator, $metadataFactory, AbstractProxyFactory::AUTOGENERATE_FILE_NOT_EXISTS_OR_CHANGED]
+        );
+
+        $proxyFactory
+            ->expects($this->any())
```
greg0ire: This line does nothing and can be dropped entirely.
```suggestion
```

tests/Doctrine/Tests/Common/Proxy/AbstractProxyFactoryTest.php
```diff
@@ -215,6 +216,170 @@ public function testMissingPrimaryKeyValue()
 
         $proxyFactory->getProxy('Class', []);
     }
+
+    public function testGetProxyFileWhenProxyDoesNotExist()
```
greg0ire:
```suggestion
    public function testGetProxyFileWhenProxyDoesNotExist(): void
```

tests/Doctrine/Tests/Common/Proxy/AbstractProxyFactoryTest.php
```diff
@@ -215,6 +216,170 @@ public function testMissingPrimaryKeyValue()
...
+            ->expects($this->any())
+            ->method('createProxyDefinition')
+            ->will($this->returnValue($definition));
+
+        $generatedProxy = $proxyFactory->getProxy('Class', ['id' => 1]);
+
+        self::assertInstanceOf('MyObject1', $generatedProxy);
+    }
+
+    public function testGetProxyFileWhenProxyIsOlderThanSource()
```
greg0ire:
```suggestion
    public function testGetProxyFileWhenProxyIsOlderThanSource(): void
```

tests/Doctrine/Tests/Common/Proxy/AbstractProxyFactoryTest.php
```diff
@@ -215,6 +216,170 @@ public function testMissingPrimaryKeyValue()
...
+            ->method('generateProxyClass');
+
+        /** @var MockObject&AbstractProxyFactory $proxyFactory */
+        $proxyFactory = $this->getMockForAbstractClass(
+            AbstractProxyFactory::class,
+            [$proxyGenerator, $metadataFactory, AbstractProxyFactory::AUTOGENERATE_FILE_NOT_EXISTS_OR_CHANGED]
+        );
+
+        $proxyFactory
+            ->expects($this->any())
```
greg0ire:
```suggestion
```

tests/Doctrine/Tests/Common/Proxy/AbstractProxyFactoryTest.php
```diff
@@ -215,6 +216,170 @@ public function testMissingPrimaryKeyValue()
...
+            ->willReturn($sourceFile);
+
+        $proxyGenerator
+            ->expects($this->once())
+            ->method('getProxyFileName')
+            ->willReturn($proxyFile);
+
+        $proxyGenerator
+            ->expects($this->once())
+            ->method('generateProxyClass');
```
greg0ire, line 307: Lots of mocking above. Can you please look around for real implementations that you could use instead? It make the test longer than it needs to be IMO. Also, I think the calls to `expects` can be dropped, because that's not really what you want to test here. That would make the above objects stubs instead of mocks, so you may try using `createStub` instead of `createMock` here (although I don't guarantee it will work, that depends on whether all builds use a recent enough version of PHPUnit)

tests/Doctrine/Tests/Common/Proxy/AbstractProxyFactoryTest.php
```diff
@@ -215,6 +216,170 @@ public function testMissingPrimaryKeyValue()
...
+
+        $generatedProxy = $proxyFactory->getProxy('Class', ['id' => 1]);
+
+        self::assertInstanceOf('MyObject2', $generatedProxy);
+    }
+
+    public function testGetProxyFileWhenProxyIsNewerThanSource()
+    {
+        $sourceFile = tempnam(sys_get_temp_dir(), 'source');
+        sleep(1);
```
greg0ire, line 327: IMO it's not such a big deal. If it becomes one in the future, we should consider introducing something like `lcobucci/clock`.

tests/Doctrine/Tests/Common/Proxy/AbstractProxyFactoryTest.php
ruudk: Honestly, I would never do this myself. The reason why I did is because this whole test does it like that. I'd rather have a more functional tests instead. Should be possible?

ruudk: Was able to use `createStub` in a few places and remove one or 2 `expects`. 
Not able to reduce it more. Any tips?

ruudk: Too bad PHPUnit that is used doesn't support it. So I'll have to undo that

greg0ire: Yeah, just looked into it and everything is either an interface or not an easy to instantiate class, so I'm afraid you won't be able to reduce it more :disappointed: 
Functional tests would probably be better here, but that's a topic that is orthogonal to your PR.

greg0ire (approved):

ruudk:
@greg0ire Is there anything else needed for this to get merged? :)

greg0ire:
Let's try to get more reviews :)

SenseException (approved):

ruudk:
Thanks for merging this. 🎉 

Do you have an idea when this can be tagged as a new minor release? Because when it is, I can create a PR for DoctrineBundle to add support for this 😊

greg0ire:
If everything goes well, in a few minutes :slightly_smiling_face:

ruudk:
Created a PR for DoctrineBundle here https://github.com/doctrine/DoctrineBundle/pull/1407

PR #946, opened 2021-10-21
fallback getFileName() to declaring class

biji:
When `Entity` source encoded with ioncube, `$method->getFileName()` returns `false`. This error occur when generating proxy: `Warning: file(): Filename cannot be empty` 

Reference: https://github.com/doctrine/common/commit/59374594248862ccfb418bbb5fc2cf91c5ef8dd0


stof:
passing a class name to the `file()` function does not make sense.

PR #950, opened 2021-12-21
PHPStan 1.2.0

derrabus:


.github/workflows/static-analysis.yml
```diff
@@ -5,13 +5,13 @@ on:
...
       - "*.x"
-      - "master"
 
 jobs:
   static-analysis:
     name: "Static Analysis"
-    uses: "greg0ire/.github/.github/workflows/static-analysis.yml@1.1.1"
+    uses: "doctrine/.github/.github/workflows/static-analysis.yml@1.4.0"
+    with:
+      php-version: "7.4"
```
malarzm, line 17: We could remove some files from PHPStan's exclusion list if we run analysis with PHP 8.1 :)

malarzm (approved):

.github/workflows/static-analysis.yml
derrabus: Yes, but then again, I'm not eager to analyze fixtures for the Proxy tests. 😉

PR #949, opened 2021-12-21
Handle PHP 8.1 types in proxies

malarzm:
This allows using `never` and intersection types in proxied classes. Fixes #948

PR #952, opened 2021-12-26
Stop passing GITHUB_TOKEN secret explicitly

greg0ire:
Doing so now results in the following error message:
> parsing called workflow
> "doctrine/.github/.github/workflows/release-on-milestone-closed.yml@1.4.0":
> workflow is invalid: secret name `GITHUB_TOKEN` within `workflow_call`
> can not be used since it would collide with system reserved name

Moreover, the documentation states the following:

> When a reusable workflow is triggered by a caller workflow, the github
> context is always associated with the caller workflow. The called
> workflow is automatically granted access to github.token and
> secrets.GITHUB_TOKEN. For more information about the github context, see
> "Context and expression syntax for GitHub Actions."

### DIFF
--- a/.github/workflows/release-on-milestone-closed.yml
+++ b/.github/workflows/release-on-milestone-closed.yml
@@ -10,7 +10,6 @@ jobs:
     name: "Git tag, release & create merge-up PR"
     uses: "doctrine/.github/.github/workflows/release-on-milestone-closed.yml@1.1.1"
     secrets:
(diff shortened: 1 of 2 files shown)


greg0ire:
Apparently, https://github.com/doctrine/common/pull/949 should have been red, but for some reason, Static Analysis jobs did not run on that PR :thinking:

PR #955, opened 2022-01-28
Bugfix: Enum default values require a use statement for correct serialization

### LINKED ISSUE #9371: Proxy generator dumps incorrect default value for enums into lazy properties defaults
### Bug Report

|    Q        |   A
|------------ | ------
| BC Break    | no
| Version     | 2.11.0

#### Summary

When updating an app to ORM 2.11 and changing an existing string field to use enums, when the field has a default value, the proxy is incorrectly generated and results in errors similar to "ReflectionException: Class "App\Entity\Card\App\Suit" does not exist."  This is because the proxy class doesn't have the leading slash to make it a FQCN, so it's treated as a relative namespace when accessed later.

#### Current behavior

The proxy class' `$lazyPropertiesDefaults` array dumps defaults with a class name that's not fully qualified, resulting in a value that's perceived as a relative namespace path.

#### How to reproduce

- Clone https://github.com/mbabker/proxy-generator-bug
- Run `composer install`
- Run `vendor/bin/doctrine orm:generate-proxies`
- Inspect the generated proxy in the `proxies` folder

#### Expected behavior

The `$lazyPropertiesDefaults` array should use fully qualified class names to avoid class not found errors, or if it cannot support enums with lazy loading, the proxy generator skips the field.

As a temporary workaround, I can move these default assignments to the class constructor.

beberlei:
This fixes https://github.com/doctrine/orm/issues/9371


kimhemsoe (approved):

lib/Doctrine/Common/Proxy/ProxyGenerator.php
```diff
@@ -381,6 +385,42 @@ private function generateNamespace(ClassMetadata $class)
...
+            if (array_key_exists($name, $defaultProperties) && $defaultProperties[$name] instanceof BackedEnum) {
+                $enumClasses[] = get_class($defaultProperties[$name]);
+            }
+        }
+
+        return implode(
+            "\n",
+            array_map(
+                static function ($className) {
+                    return 'use ' . $className . ';';
```
kimhemsoe, line 418: Will this work if it is a deep namespace? Ex var_export export class as 'Doctrine\Tests\Common\Proxy\YesNo' So should the "use" not only be "use Doctrine"?

kimhemsoe (changes requested):

phpstan.neon.dist
```diff
@@ -1,5 +1,4 @@
 parameters:
-    phpVersion: 70100
```
derrabus, line 2: I'd keep the setting and bump it to `80100`.

lib/Doctrine/Common/Proxy/ProxyGenerator.php
beberlei: Yes, see the test assertion, https://github.com/doctrine/common/pull/955/files#diff-8bd50f2d6a0c9d08df7a09474c0ce3ddb0222b38b85984b1b723bfec5825e768R544

tests/Doctrine/Tests/Common/Proxy/ProxyGeneratorTest.php
```diff
@@ -525,6 +525,28 @@ public function testPhp81NeverType()
...
+     * @requires PHP >= 8.1.0
+     */
+    public function testEnumDefaultInPublicProperty() : void
+    {
+        $className = Php81EnumPublicPropertyType::class;
+
+        if ( ! class_exists('Doctrine\Tests\Common\ProxyProxy\__CG__\Php81EnumPublicPropertyType', false)) {
+            $metadata = $this->createClassMetadata($className, ['id']);
+
+            $metadata->expects($this->any())->method('hasField')->will($this->returnValue(true));
```
greg0ire: `->expects($this->any()` is a no-op

```suggestion
            $metadata->method('hasField')->will($this->returnValue(true));
```

phpstan.neon.dist
beberlei: Updated this, but not sure its a good idea because its configured in two locations now, in the github ci yml setup and here. This might lead to missmatches when future changes are done.

beberlei: @greg0ire whats your opinion on this?

derrabus: In case of a mismatch, this setting wins. Setting the version explicitly in this config file should make the PHPStan result reproducible in a contributor's local setup. At least I would hope so. 😅 🤞

derrabus (dismissed):

phpstan.neon.dist
greg0ire: The thing configured on Github action in the php version to install on the machine, whereas this represents the target version, I'd say these are 2 orthogonal settings, and when not specified, here, it will default to the current php version. When php 8.2 is going to come out, we still want to analyze the code as if it was written for 8.1, but might benefit from faster runs/more up to date deps by installing php 8.2. So I'd say it's a good practice to specify it.

greg0ire: I'm not sure what happens if we have bits of code that uses syntax specific to 8.2 though :thinking: (like we have test code that uses `readonly` properties in the ORM).

SenseException (dismissed):

lib/Doctrine/Common/Proxy/ProxyGenerator.php
kimhemsoe: I did not write what i meant clearly enough.

Try "new \Doctrine\Tests\Common\ProxyProxy\__CG__\Doctrine\Tests\Common\Proxy\Php81EnumPublicPropertyType()" and you will get "Error: Class "Doctrine\Tests\Common\ProxyProxy\__CG__\Doctrine\Tests\Common\Proxy\Doctrine\Tests\Common\Proxy\YesOrNo" not found"

I believe that we only should "use Doctrine" and not "use Doctrine\Tests\Common\Proxy\YesOrNo"

kimhemsoe: A quick and dirty example of a fix:

$rootNamespaces = array_map(fn ($enumClass) => explode('\\', $enumClass)[0], $enumClasses);

kimhemsoe (changes requested):

kimhemsoe (approved):

PR #957, opened 2022-02-11
Don't Generate Set or Unset on Proxy of Readonly Public Properties

### LINKED ISSUE #954: Public readonly properties are unset in Proxy's constructor
### Failing Test

|    Q        |   A
|------------ | ------
| BC Break    | no
| Version     | 3.2.0

- https://github.com/doctrine/orm/issues/9432
- https://github.com/doctrine/orm/pull/9431

#### Summary

Public readonly properties are unset in Proxy's constructor, which results in error.

#### Current behavior

`Cannot unset readonly property \Entity\SimpleBook::$title from scope Proxies\__CG__\Entity\SimpleBook`

#### How to reproduce

```php
#[Entity, Table]
class SimpleBook
{
    #[Column, Id, GeneratedValue]
    private readonly int $id;

    #[Column]
    public readonly string $title;

    #[ManyToOne, JoinColumn(nullable: false)]
    private readonly Author $author;

    public function getId(): int
    {
        return $this->id;
    }

    public function getTitle(): string
    {
        return $this->title;
    }

    public function getAuthor(): Author
    {
        return $this->author;
    }
}

class SimpleBook extends \Entity\SimpleBook implements \Doctrine\ORM\Proxy\Proxy
{
    // ...

    public function __construct(?\Closure $initializer = null, ?\Closure $cloner = null)
    {
        unset($this->title);

        $this->__initializer__ = $initializer;
        $this->__cloner__      = $cloner;
    }
}
```

#### Code that generates Proxy's constructor

I dug into ProxyGenerator and here is generation of these unsets:

- https://github.com/doctrine/common/blob/3.2.x/lib/Doctrine/Common/Proxy/ProxyGenerator.php#L979

`public readonly ` properties are not tested as I see in tests.

- https://github.com/doctrine/common/blob/3.2.x/tests/Doctrine/Tests/Common/Proxy/ProxyGeneratorTest.php#L481

zanbaldwin:
Attempting to fix doctrine/common#954.

- Create another private method `getWriteableLazyLoadedPublicPropertiesNames()`
- Use this method when generating `__set` or generating calls to `unset()` in `__construct` and `__wakeup`
- Leave all other methods that referenced `getLazyLoadedPublicPropertiesNames()` as they were.

Would like input on these changes as I don't have enough knowledge to know if omitting these properties will have unintended side-effects elsewhere.

### DIFF
--- a/lib/Doctrine/Common/Proxy/ProxyGenerator.php
+++ b/lib/Doctrine/Common/Proxy/ProxyGenerator.php
@@ -1017,7 +1017,8 @@ EOT;
         foreach ($class->getReflectionClass()->getProperties(ReflectionProperty::IS_PUBLIC) as $property) {
             $name = $property->getName();
 
-            if ((! $class->hasField($name) && ! $class->hasAssociation($name))
+            if (
+                (! $class->hasField($name) && ! $class->hasAssociation($name))
                 || $class->isIdentifier($name)
                 || (method_exists($property, 'isReadOnly') && $property->isReadOnly())
             ) {
--- a/lib/Doctrine/Common/Proxy/ProxyGenerator.php
+++ b/lib/Doctrine/Common/Proxy/ProxyGenerator.php
@@ -474,7 +474,7 @@ EOT;
 
         $toUnset = array_map(static function (string $name): string {
             return '$this->' . $name;
-        }, $this->getLazyLoadedPublicPropertiesNames($class));
+        }, $this->getWriteableLazyLoadedPublicPropertiesNames($class));
 
         return $constructorImpl . ($toUnset === [] ? '' : '        unset(' . implode(', ', $toUnset) . ");\n")
             . <<<'EOT'
@@ -591,7 +591,7 @@ EOT
      */
     private function generateMagicSet(ClassMetadata $class)
     {
-        $lazyPublicProperties = $this->getLazyLoadedPublicPropertiesNames($class);
+        $lazyPublicProperties = $this->getWriteableLazyLoadedPublicPropertiesNames($class);
         $reflectionClass      = $class->getReflectionClass();
         $hasParentSet         = false;
         $inheritDoc           = '';
@@ -808,7 +808,7 @@ EOT;
         $hasParentWakeup = $reflectionClass->hasMethod('__wakeup');
 
         $unsetPublicProperties = [];
-        foreach ($this->getLazyLoadedPublicPropertiesNames($class) as $lazyPublicProperty) {
+        foreach ($this->getWriteableLazyLoadedPublicPropertiesNames($class) as $lazyPublicProperty) {
             $unsetPublicProperties[] = '$this->' . $lazyPublicProperty;
         }
 
@@ -1003,6 +1003,31 @@ EOT;
         }
 
         return false;
+    }
+
+    /**
+     * Generates the list of public properties to be lazy loaded, that are writable.
+     *
+     * @return array<int, string>
+     */
+    public function getWriteableLazyLoadedPublicPropertiesNames(ClassMetadata $class): array
+    {
+        $properties = [];
+
+        foreach ($class->getReflectionClass()->getProperties(ReflectionProperty::IS_PUBLIC) as $property) {
+            $name = $property->getName();
+
+            if ((! $class->hasField($name) && ! $class->hasAssociation($name))
+                || $class->isIdentifier($name)
+                || (method_exists($property, 'isReadOnly') && $property->isReadOnly())
+            ) {
+                continue;
+            }
+
+            $properties[] = $name;
+        }
+
+        return $properties;
     }
 
     /**
--- a/lib/Doctrine/Common/Proxy/ProxyGenerator.php
+++ b/lib/Doctrine/Common/Proxy/ProxyGenerator.php
@@ -1008,7 +1008,7 @@ EOT;
     /**
      * Generates the list of public properties to be lazy loaded, that are writable.
      *
-     * @return array<int, string>
+     * @return list<string>
      */
     public function getWriteableLazyLoadedPublicPropertiesNames(ClassMetadata $class): array
     {
--- a/tests/Doctrine/Tests/Common/Proxy/ProxyGeneratorTest.php
+++ b/tests/Doctrine/Tests/Common/Proxy/ProxyGeneratorTest.php
@@ -528,6 +528,41 @@ class ProxyGeneratorTest extends TestCase
     /**
      * @requires PHP >= 8.1.0
      */
+    public function testPhp81ReadonlyPublicProperties()
+    {
+        $className = Php81ReadonlyPublicPropertyType::class;
+        $proxyClassName = 'Doctrine\Tests\Common\ProxyProxy\__CG__\Php81ReadonlyPublicPropertyType';
+
+        if ( ! class_exists($proxyClassName, false)) {
+            $metadata = $this->createClassMetadata($className, ['id']);
+
+            $metadata
+                ->expects($this->any())
+                ->method('hasField')
+                ->will($this->returnCallback(static function ($fieldName) {
+                    return in_array($fieldName, ['id', 'readable', 'writeable']);
+                }));
+
+            $proxyGenerator = new ProxyGenerator(__DIR__ . '/generated', __NAMESPACE__ . 'Proxy');
+            $this->generateAndRequire($proxyGenerator, $metadata);
+        }
+
+        // Readonly properties are removed from unset.
+        self::assertStringContainsString(
+            'unset($this->writeable);',
+            file_get_contents(__DIR__ . '/generated/__CG__DoctrineTestsCommonProxyPhp81ReadonlyPublicPropertyType.php')
+        );
+
+        // But remain in property listings.
+        self::assertStringContainsString(
+            "'readable' => NULL",
+            file_get_contents(__DIR__ . '/generated/__CG__DoctrineTestsCommonProxyPhp81ReadonlyPublicPropertyType.php')
+        );
+    }
+
+    /**
+     * @requires PHP >= 8.1.0
+     */
     public function testEnumDefaultInPublicProperty() : void
     {
         $className = Php81EnumPublicPropertyType::class;
--- a/tests/Doctrine/Tests/Common/Proxy/ProxyGeneratorTest.php
+++ b/tests/Doctrine/Tests/Common/Proxy/ProxyGeneratorTest.php
@@ -4,6 +4,7 @@ namespace Doctrine\Tests\Common\Proxy;
 
 use Doctrine\Common\Proxy\Exception\InvalidArgumentException;
 use Doctrine\Common\Proxy\Exception\UnexpectedValueException;
+use Doctrine\Common\Proxy\Proxy;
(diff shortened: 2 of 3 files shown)


lib/Doctrine/Common/Proxy/ProxyGenerator.php
```diff
@@ -1005,6 +1005,31 @@ private function isShortIdentifierGetter($method, ClassMetadata $class)
         return false;
     }
 
+    /**
+     * Generates the list of public properties to be lazy loaded, that are writable.
+     *
+     * @return array<int, string>
```
greg0ire:
```suggestion
     * @return list<string>
```

greg0ire (commented):
I think that for the test, you would need to create a class with a readonly property, then generate a proxy for it, then instantiate that proxy.

Here is an existing test that seems to do that:

https://github.com/doctrine/common/blob/295082d3750987065912816a9d536c2df735f637/tests/Doctrine/Tests/Common/Proxy/ProxyMagicMethodsTest.php#L80-L106

Once you have that, you can validate that the bug no longer happens :slightly_smiling_face:

zanbaldwin:
Thanks for the suggestions! :slightly_smiling_face:

tests/Doctrine/Tests/Common/Proxy/ProxyGeneratorTest.php
```diff
@@ -525,6 +525,41 @@ public function testPhp81NeverType()
...
+        self::assertStringContainsString(
+            'unset($this->writeable);',
+            file_get_contents(__DIR__ . '/generated/__CG__DoctrineTestsCommonProxyPhp81ReadonlyPublicPropertyType.php')
+        );
+
+        // But remain in property listings.
+        self::assertStringContainsString(
+            "'readable' => NULL",
+            file_get_contents(__DIR__ . '/generated/__CG__DoctrineTestsCommonProxyPhp81ReadonlyPublicPropertyType.php')
+        );
```
greg0ire: I think it would be great to do a `new $proxyClassName` here and perform more assertions.

zanbaldwin: I'm at the point where I don't know how to proceed, I've tried setting data inside the proxy initialization proxy using the following:

```php
$initializationData = [
    'id' => 'c0b5cb93-f01b-43f8-bc66-bc943b1ebcfd',
    'readable' => 'This field is read-only.',
    'writeable' => 'This field is writeable.',
];
$proxy = new $proxyClassName(static function (Proxy $proxy, $method, $params) use (&$counter, $initializationData) {
    if (!in_array($params[0], array_keys($initializationData))) {
        throw new InvalidArgumentException(
            sprintf('Should not be initialized when checking isset("%s")', $params[0])
        );
    }
    $initializer = $proxy->__getInitializer();
    $proxy->__setInitializer(null);
    $proxy->{$params[0]} = $initializationData[$params[0]];
    $counter += 1;
    $proxy->__setInitializer($initializer);
});
self::assertTrue(isset($proxy->id));
```

Not sure how to initialize a readonly property from the proxy initializer closure, when any readonly property initialization is required to happen from inside the class itself.

There must be a way since ORM supports readonly properties, but I've reached my limit for tonight as I found another potential bug while diving into the UnitOfWork internals.

greg0ire: Maybe you can define another closure inside the first one, but this time, bind it to the proxy? Like this: https://3v4l.org/Q1rTq#v8.1.2

zanbaldwin: That was a good idea, but the initializer closure is never getting called because PHP core is throwing an error (trying to access uninitialized property) before any magic methods are called. I'm officially stumped.

greg0ire: That's confusing indeed! Do you have a stack trace with that error?

zanbaldwin: A `print_r` of the error thrown by the above code, which isn't very helpful - it looks like an error is thrown as soon as it gets to trying to access the property and doesn't proceed any further:

```
Error Object
(
    [message:protected] => Typed property Doctrine\Tests\Common\Proxy\Php81ReadonlyPublicPropertyType::$id must not be accessed before initialization
    [string:Error:private] =>
    [code:protected] => 0
    [file:protected] => /doctrine-common/tests/Doctrine/Tests/Common/Proxy/ProxyGeneratorTest.php
    [line:protected] => 575
    [trace:Error:private] => (replaced with Throwable::getTraceAsString)
        #0 /doctrine-common/vendor/phpunit/phpunit/src/Framework/TestCase.php(1545): Doctrine\Tests\Common\Proxy\ProxyGeneratorTest->testPhp81ReadonlyPublicProperties()
        #1 /doctrine-common/vendor/phpunit/phpunit/src/Framework/TestCase.php(1151): PHPUnit\Framework\TestCase->runTest()
        #2 /doctrine-common/vendor/phpunit/phpunit/src/Framework/TestResult.php(726): PHPUnit\Framework\TestCase->runBare()
        #3 /doctrine-common/vendor/phpunit/phpunit/src/Framework/TestCase.php(903): PHPUnit\Framework\TestResult->run()
        #4 /doctrine-common/vendor/phpunit/phpunit/src/Framework/TestSuite.php(677): PHPUnit\Framework\TestCase->run()
        #5 /doctrine-common/vendor/phpunit/phpunit/src/Framework/TestSuite.php(677): PHPUnit\Framework\TestSuite->run()
        #6 /doctrine-common/vendor/phpunit/phpunit/src/Framework/TestSuite.php(677): PHPUnit\Framework\TestSuite->run()
        #7 /doctrine-common/vendor/phpunit/phpunit/src/TextUI/TestRunner.php(670): PHPUnit\Framework\TestSuite->run()
        #8 /doctrine-common/vendor/phpunit/phpunit/src/TextUI/Command.php(143): PHPUnit\TextUI\TestRunner->run()
        #9 /doctrine-common/vendor/phpunit/phpunit/src/TextUI/Command.php(96): PHPUnit\TextUI\Command->run()
        #10 /doctrine-common/vendor/phpunit/phpunit/phpunit(98): PHPUnit\TextUI\Command::main()
        #11 /doctrine-common/vendor/bin/phpunit(110): include('...')
        #12 {main}
    [previous:Error:private] =>
)
```

I would ask if my proxy initializer closure is correct, but I don't think it's even getting to that point:

```php
$proxy = new $proxyClassName(static function (Proxy $proxy, $method, $params) {
    echo 'Proxy initialization closure called for property "' . $params[0] . '".';
    exit;
});

var_dump(isset($proxy->id)); // will always return false.
var_dump($proxy->id)         // throws Error.
```

zanbaldwin: May have to pass in initial values for the readonly properties into the proxy constructor, which kinda defeats the whole point of a proxy.

tests/Doctrine/Tests/Common/Proxy/ProxyGeneratorTest.php
```diff
@@ -525,6 +526,72 @@ public function testPhp81NeverType()
...
+                );
+            }
+            $initializer = $proxy->__getInitializer();
+            $proxy->__setInitializer(null);
+            isset($this->{$params[0]}) || $this->{$params[0]} = $initializationData[$params[0]];
+            $counter += 1;
+            $proxy->__setInitializer($initializer);
+        });
+
+        var_dump($proxy->id);
```
greg0ire: This should go away, right?

tests/Doctrine/Tests/Common/Proxy/ProxyGeneratorTest.php
```diff
@@ -525,6 +526,72 @@ public function testPhp81NeverType()
...
+            'id' => 'c0b5cb93-f01b-43f8-bc66-bc943b1ebcfd',
+            'readable' => 'This field is read-only.',
+            'writeable' => 'This field is writeable.',
+        ];
+
+        if ( ! class_exists($proxyClassName, false)) {
+            $metadata = $this->createClassMetadata($className, ['id']);
+
+            $metadata
+                ->expects($this->any())
```
greg0ire: I remember already commenting on this, but for some reason my comment does not appear… maybe I discarded by accident. 

You can drop this line, it's equivalent.
```suggestion
```

tests/Doctrine/Tests/Common/Proxy/ProxyGeneratorTest.php
```diff
@@ -525,6 +526,72 @@ public function testPhp81NeverType()
...
+                ->method('hasField')
+                ->will($this->returnCallback(static function ($fieldName) use ($initializationData) {
```
greg0ire:
```suggestion
                ->will($this->returnCallback(static function (string $fieldName) use ($initializationData): bool {
```

tests/Doctrine/Tests/Common/Proxy/ProxyGeneratorTest.php
```diff
@@ -525,6 +526,72 @@ public function testPhp81NeverType()
...
+
+        self::assertSame('This field is writeable.', $proxy->writeable);
+        $proxy->writeable = 'Updated string contents.';
+        self::assertSame('Updated string contents.', $proxy->writeable);
+
+        try {
+            $proxy->readable = 'Invalid';
+            self::fail('Should not be able to update readonly property.');
+        } catch (\Error) {
+        }
```
greg0ire: I didn't know you could do this, and apparently it's only possible on recent versions of PHP.

```suggestion
        self::expectError();
        $proxy->readable = 'Invalid';
```

tests/Doctrine/Tests/Common/Proxy/ProxyGeneratorTest.php
greg0ire: @beberlei @derrabus @malarzm might have some ideas here :crossed_fingers:

beberlei:
This has the unintended consequence of breaking lazy loading if the user acceses the readonly public property of an not yet loaded proxy, then it will not be loaded leaving the property NULL.

tests/Doctrine/Tests/Common/Proxy/ProxyGeneratorTest.php
chapterjason: @greg0ire @zanbaldwin I had an issue with the same topic: 

https://github.com/doctrine/orm/issues/10049
https://github.com/doctrine/orm/pull/10059

You can set properties by changing the scope of the reflection property.

alessandromorelli:
The readonly public properties should be handled separately from the other public properties.

The readonly public properties should be unset but should not be initialized with a default, to let the initializer free to write them.

I cobbled up a proof of concept:
https://github.com/alessandromorelli/doctrine-common/commit/7f9d1d0206e14e16f832883b80d113599f0dfecb

Readonly properties may be unset only in the same scope in which they are declared, so I've bound a Closure to the proxy's parent class.

Once unset, any read access to the property triggers the __get method and the initializer can insert the values.

If the initializer does not actually initialize the readonly property, any write access triggers a "Cannot initialize readonly property XXXX from scope YYYY", which is a bit confusing but shouldn't normally happen.

PR #963, opened 2022-05-29
Use random_bytes instead of uniqid

### LINKED ISSUE #962: Proxy generator hangs indefinitely when system time is mocked
# Bug Report
Proxy generator hangs indefinitely when system time is mocked

Q | A
-- | --
BC Break | no
Version | 2.x/3.x

### Summary

Because `uniqid` relies on system time, when the clock is mocked (e.g. with libfaketime) and it does not advance, calling uniqid more than 1 times will hang up the process indefinitely. This means, when running tests with mocked constant time, the code gets stuck at https://github.com/doctrine/common/blob/c824e95d4c83b7102d8bc60595445a6f7d540f96/lib/Doctrine/Common/Proxy/ProxyGenerator.php#L341

---

### How to reproduce the issue with plain`uniqid`:

```bash
docker build -t php-faketime - <<DOCKERFILE
FROM php:8.1-cli
RUN apt-get update && apt-get install -y libfaketime   
ENV LD_PRELOAD=/usr/lib/x86_64-linux-gnu/faketime/libfaketime.so.1
DOCKERFILE

docker run --rm -e FAKETIME="2000-02-02 02:02:02" php-faketime -r 'echo uniqid(), PHP_EOL, uniqid();'
``` 

Ideally this would be patched in PHP but I think that won't happen since even the [docs](https://www.php.net/manual/en/function.uniqid.php) mention that `uniqid` is using microtime as random seed, so this is probably considered as a feature and not a bug.

### Possible solutions
I think it would be better to use something like [`tempnam`](https://www.php.net/manual/en/function.tempnam.php), or some other time-independent random function.

---

Related discussion in [libfaketime repository](https://github.com/wolfcw/libfaketime/issues/349).

valtzu:
Closes #962

### DIFF
--- a/lib/Doctrine/Common/Proxy/ProxyGenerator.php
+++ b/lib/Doctrine/Common/Proxy/ProxyGenerator.php
@@ -22,6 +22,7 @@ use function array_map;
 use function array_slice;
 use function array_unique;
 use function assert;
+use function bin2hex;
 use function call_user_func;
 use function chmod;
 use function class_exists;
@@ -43,6 +44,7 @@ use function method_exists;
 use function mkdir;
 use function preg_match;
 use function preg_match_all;
+use function random_bytes;
 use function rename;
 use function rtrim;
 use function sprintf;
@@ -52,7 +54,6 @@ use function strtolower;
 use function strtr;
 use function substr;
 use function trim;
-use function uniqid;
 use function var_export;
 
 use const DIRECTORY_SEPARATOR;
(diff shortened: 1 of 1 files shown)


lib/Doctrine/Common/Proxy/ProxyGenerator.php
```diff
@@ -338,7 +339,8 @@ public function generateProxyClass(ClassMetadata $class, $fileName = false)
             throw UnexpectedValueException::proxyDirectoryNotWritable($this->proxyDirectory);
         }
 
-        $tmpFileName = $fileName . '.' . uniqid('', true);
+        /** @noinspection PhpUnhandledExceptionInspection */
```
greg0ire: No IDE-specific stuff please 🙏

valtzu: Ok, sorry, removed.

greg0ire (approved):
Maintainers : to be squash merged

derrabus (approved):

PR #966, opened 2022-08-07
Support new in initializers

### LINKED ISSUE #961: Compile Error: Constant expression contains invalid operations
### Bug Report

Compile Error: Constant expression contains invalid operations

|    Q        |   A
|------------ | ------
| BC Break    | no
| Version     | 2.11.2

#### Summary

Entity with following method
```
class A {
    public function getActivePriceOverrideForCompany(
        Company $company,
        DateTimeImmutable $activeAt = new DateTimeImmutable('today')
    ): ?PriceOverride { ... }
}
```

as result getting proxy like this:

```
public function getActivePriceOverrideForCompany(\App\Entity\Company $company, \DateTimeImmutable $activeAt = DateTimeImmutable::__set_state(array(
   'date' => '2022-05-03 00',
   'timezone_type' => 3,
   'timezone' => 'UTC',
))): ?\App\Entity\PriceOverride { ... }
```

1. Proxy is invalid
2. Even if the code will be correct, proxy will be invalid on the next day.

This will work:
```
class A {
    public function getActivePriceOverrideForCompany(
        Company $company,
        ?DateTimeImmutable $activeAt = null
    ): ?PriceOverride { ... }
}
```

malarzm:
Port of https://github.com/Ocramius/ProxyManager/pull/755 
Fixes #961

### DIFF
--- a/lib/Doctrine/Common/Proxy/ProxyGenerator.php
+++ b/lib/Doctrine/Common/Proxy/ProxyGenerator.php
@@ -1091,10 +1091,7 @@ EOT;
             }
 
             $parameterDefinition .= '$' . ($renameParameters ? $renameParameters[$i] : $param->getName());
-
-            if ($param->isDefaultValueAvailable()) {
-                $parameterDefinition .= ' = ' . var_export($param->getDefaultValue(), true);
-            }
+            $parameterDefinition .= $this->getParameterDefaultValue($param);
 
             $parameterDefinitions[] = $parameterDefinition;
         }
@@ -1116,6 +1113,24 @@ EOT;
         assert($declaringFunction instanceof ReflectionMethod);
 
         return $this->formatType($parameter->getType(), $declaringFunction, $parameter);
+    }
+
+    /**
+     * @return string
+     */
+    private function getParameterDefaultValue(ReflectionParameter $parameter)
+    {
+        if (! $parameter->isDefaultValueAvailable()) {
+            return '';
+        }
+
+        if (PHP_VERSION_ID < 80100) {
+            return ' = ' . var_export($parameter->getDefaultValue(), true);
+        }
+
+        $value = rtrim(substr(explode('$' . $parameter->getName() . ' = ', (string) $parameter, 2)[1], 0, -2));
+
+        return ' = ' . $value;
     }
 
     /**
--- a/tests/Doctrine/Tests/Common/Proxy/PHP81NewInInitializers.php
+++ b/tests/Doctrine/Tests/Common/Proxy/PHP81NewInInitializers.php
@@ -1,0 +1,23 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\Tests\Common\Proxy;
+
+class PHP81NewInInitializers
+{
+    public function onlyInitializer($foo = new \stdClass()): void
+    {
+
+    }
+
+    public function typed(\DateTimeInterface $foo = new \DateTimeImmutable('now')): void
+    {
+
+    }
+
+    public function arrayInDefault(array $foo = [new \DateTimeImmutable('2022-08-22 16:20', new \DateTimeZone('Europe/Warsaw'))]): void
+    {
+
+    }
+}
--- a/tests/Doctrine/Tests/Common/Proxy/ProxyGeneratorTest.php
+++ b/tests/Doctrine/Tests/Common/Proxy/ProxyGeneratorTest.php
@@ -573,6 +573,36 @@ class ProxyGeneratorTest extends TestCase
     }
 
     /**
+     * @requires PHP >= 8.1.0
+     */
+    public function testPhp81NewInInitializers()
+    {
+        $className = PHP81NewInInitializers::class;
+
+        if (!class_exists('Doctrine\Tests\Common\ProxyProxy\__CG__\PHP81NewInInitializers', false)) {
+            $metadata = $this->createClassMetadata($className, ['id']);
+
+            $proxyGenerator = new ProxyGenerator(__DIR__ . '/generated', __NAMESPACE__ . 'Proxy');
+            $this->generateAndRequire($proxyGenerator, $metadata);
+        }
+
+        self::assertStringContainsString(
+            'onlyInitializer($foo = new \stdClass()): void',
+            file_get_contents(__DIR__ . '/generated/__CG__DoctrineTestsCommonProxyPHP81NewInInitializers.php')
+        );
+
+        self::assertStringContainsString(
+            'typed(\DateTimeInterface $foo = new \DateTimeImmutable(\'now\')): void',
+            file_get_contents(__DIR__ . '/generated/__CG__DoctrineTestsCommonProxyPHP81NewInInitializers.php')
+        );
+
+        self::assertStringContainsString(
+            'arrayInDefault(array $foo = [new \DateTimeImmutable(\'2022-08-22 16:20\', new \DateTimeZone(\'Europe/Warsaw\'))]): void',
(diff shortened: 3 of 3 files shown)


derrabus:
The static analysis failure is my fault, I guess. 🙈 See #967.

malarzm:
@derrabus I knew I've seen that PR somewhere, couldn't recall where :D thanks for creating a fix so quickly 🍻

derrabus:
Please rebase, static analysis should be green now. 🤞🏻

derrabus:
@nicolas-grekas Can you have a look, please?

lib/Doctrine/Common/Proxy/ProxyGenerator.php
```diff
@@ -1116,6 +1116,33 @@ private function getParameterType(ReflectionParameter $parameter)
         return $this->formatType($parameter->getType(), $declaringFunction, $parameter);
     }
 
+    /**
+     * @return string
```
greg0ire, line 1119: Why not use a native type declaration here?

lib/Doctrine/Common/Proxy/ProxyGenerator.php
```diff
@@ -1116,6 +1116,33 @@ private function getParameterType(ReflectionParameter $parameter)
...
+
+        $value = rtrim(substr(explode('$' . $parameter->getName() . ' = ', (string) $parameter, 2)[1], 0, -2));
+        $parts = preg_split('{(\'(?:[^\'\\\\]*(?:\\\\.[^\'\\\\]*)*)\')}', $value, -1, PREG_SPLIT_DELIM_CAPTURE);
+
+        foreach ($parts as $i => &$part) {
+            if ($part === '' || $i % 2 !== 0) {
+                continue;
+            }
+
+            $part = preg_replace('/(?(DEFINE)(?<V>[a-zA-Z_\x7f-\xff][a-zA-Z0-9_\x7f-\xff]*+))(?<!\\\\)(?&V)(?:\\\\(?&V))*+::/', '\\\\$0', $part);
```
greg0ire: What is the purpose of this line? When I comment it out, the test below does not fail :thinking: … I'm using PHP 8.1.8 Likewise, if I put rubbish in the regex line 1133, no test fails.

lib/Doctrine/Common/Proxy/ProxyGenerator.php
```diff
@@ -1116,6 +1116,33 @@ private function getParameterType(ReflectionParameter $parameter)
...
+        if (! $parameter->isDefaultValueAvailable()) {
+            return '';
+        }
+
+        if (PHP_VERSION_ID < 80100) {
+            return ' = ' . var_export($parameter->getDefaultValue(), true);
+        }
+
+        $value = rtrim(substr(explode('$' . $parameter->getName() . ' = ', (string) $parameter, 2)[1], 0, -2));
+        $parts = preg_split('{(\'(?:[^\'\\\\]*(?:\\\\.[^\'\\\\]*)*)\')}', $value, -1, PREG_SPLIT_DELIM_CAPTURE);
```
greg0ire: It would be nice to use the `x` delimiter and add comments inside that regex. If that makes it become multiline, then maybe also use nowdoc? That way there won't be a need for escaping the single quotes.

lib/Doctrine/Common/Proxy/ProxyGenerator.php
malarzm: I have absolutely no clue, I took what @nicolas-grekas provided for ProxyManager for granted 🙈 Tried to go for original FriendsOfPHP PR but nobody challenged that there either

lib/Doctrine/Common/Proxy/ProxyGenerator.php
malarzm: Same thing as above

lib/Doctrine/Common/Proxy/ProxyGenerator.php
malarzm: No particular reason, I was mimicking other methods in the generator.

lib/Doctrine/Common/Proxy/ProxyGenerator.php
greg0ire: Then let's just return what's on line 1132? It seems to be the important part of this patch. When somebody reports the bug the bottom part fixes, let's make sure to add a test that documents it, as well as comments.

nicolas-grekas: This regexp is supposed to add the `\` prefix to internal classes when they declare a default value that references a class constant.

greg0ire: This sounds like it would also affect top-level classes that are not internal, but user defined, wouldn't it? In both cases, I'm unsure if `doctrine/common` should support it, as AFAIK it's only ever used to proxify entities and documents, that should never be in the top-level namespace, or inherit from a class defined in it. Unlike proxy-manager I suppose.

nicolas-grekas: User defined classes always have the backslash prefix that's why only internal ones are concerned.
Agree about entities, unless they could extend e.g. ArrayObject or similar? Not sure it'd make sense.

greg0ire: That would be quite weird but I not completely impossible I suppose. @malarzm , I think we should either add a test case based on `ArrayObject::asort()` (which refers to a constant), or drop the code.

malarzm: Added part with using global const, it does fail without those regexes :)

lib/Doctrine/Common/Proxy/ProxyGenerator.php
malarzm: Let's deal with those globally later? We will be able to add them to private methods only though

lib/Doctrine/Common/Proxy/ProxyGenerator.php
greg0ire: Really? When I try again what I did in https://github.com/doctrine/common/pull/966#discussion_r947094561 on eb0aaa523902d77fcc8cbccf06fa7f4f6fd869f3, the test still passes :thinking: 

Also, dumping `$part` shows no prepended backslash.

malarzm: > Really?

Dang, sorry. I made an early return in `getDefaultValue` that was supposed to break the test... it did, but for totally other reason :s

lib/Doctrine/Common/Proxy/ProxyGenerator.php
greg0ire: `doctrine/common` is probably going to die soon, so that's really not a big deal, we can deal with it later or not at all.

lib/Doctrine/Common/Proxy/ProxyGenerator.php
malarzm: Sadly, I'm giving up. I'm almost starting my vacations and I think it'll be better to have main cases covered. If somebody reports a bug that could be fixed by regexes I've just removed then it can be done in a bugfix manner.

greg0ire (approved):
Have a nice vacation!

lib/Doctrine/Common/Proxy/ProxyGenerator.php
rvanlaak: Interesting! @greg0ire are you referring to https://github.com/doctrine/common/issues/826 here?

greg0ire: I'm referring to a discussion on a private Slack.

PR #969, opened 2022-08-08
Fix PHPUnit deprecations

malarzm:
As per title :)

### DIFF
--- a/tests/Doctrine/Tests/Common/Proxy/ProxyLogicTest.php
+++ b/tests/Doctrine/Tests/Common/Proxy/ProxyLogicTest.php
@@ -198,8 +198,14 @@ class ProxyLogicTest extends TestCase
         $this->configureInitializerMock(0);
 
         $class = get_class($this->lazyObject);
-        $this->expectException(Notice::class);
-        $this->expectExceptionMessage('Undefined property: ' . $class . '::$non_existing_property');
+        // @todo drop condition when PHPUnit 9.x becomes lowest
+        if (method_exists($this, 'expectNotice')) {
+            $this->expectNotice();
+            $this->expectNoticeMessage('Undefined property: ' . $class . '::$non_existing_property');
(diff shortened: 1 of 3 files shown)


tests/Doctrine/Tests/Common/Proxy/ProxyLogicTest.php
```diff
@@ -193,13 +192,16 @@ static function () use ($test) {
         self::assertSame('publicAssociationValue', $this->lazyObject->publicAssociation);
     }
 
+    /**
+     * @requires PHP > 7.1
```
malarzm, line 196: `expectNotice` does not exist in PHPUnit running on 7.1. I figured the test can be skipped as this is tested in higher versions anyway but I can also check for `expectNotice` existence and conditionally call old method.

tests/Doctrine/Tests/Common/Proxy/ProxyLogicTest.php
```diff
@@ -193,13 +192,16 @@ static function () use ($test) {
         self::assertSame('publicAssociationValue', $this->lazyObject->publicAssociation);
     }
 
+    /**
+     * @requires PHP >= 7.2
```
malarzm: `expectNotice` does not exist in PHPUnit running on 7.1. I figured the test can be skipped as this is tested in higher versions anyway but I can also check for `expectNotice` existence and conditionally call old method.

derrabus: > I can also check for `expectNotice` existence and conditionally call old method.

I'd prefer that. You could also polyfill `expectNotice` with a trait.

derrabus:
For 3.3.x since it affects the test suite only?

derrabus (approved):

derrabus:
Thanks!

PR #970, opened 2022-08-19
Support void return type for __clone magic method

deguif:
Starting PHP 8.0, `void` an be type-hinted as return for `__clone` method.
But proxy generated doesn't add this return type, which breaks the inherited contract.


greg0ire (approved):

malarzm (approved):

malarzm:
Thanks @deguif!

deguif:
@greg0ire @malarzm thanks for the quick review, merge and release.

PR #973, opened 2022-08-23
Remove reference to version file

greg0ire:
It has been removed in #893


stof (approved):

stof:
`make-release` in the ant config file (build.xml) was relying on that version file. I think the ant tooling should be removed entirely.

PR #977, opened 2022-08-23
Remove useless ignore rules

greg0ire:
The referenced issue is fixed, but the fix is not released yet. However,
since the "tests" directory is not currently analyzed, the ignore rule
is useless.

Also, the rule about ignoring the vendor directory is useless.

### DIFF
--- a/psalm.xml
+++ b/psalm.xml
@@ -10,10 +10,7 @@
     <projectFiles>
         <directory name="lib/Doctrine/Common" />
         <ignoreFiles>
-            <!-- Remove ProxyGeneratorTest once Psalm supports native intersection https://github.com/vimeo/psalm/issues/6413 -->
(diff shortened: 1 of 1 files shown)


psalm.xml
```diff
@@ -10,9 +10,7 @@
     <projectFiles>
         <directory name="lib/Doctrine/Common" />
         <ignoreFiles>
-            <!-- Remove ProxyGeneratorTest once Psalm supports native intersection https://github.com/vimeo/psalm/issues/6413 -->
             <file name="lib/Doctrine/Common/Proxy/ProxyGenerator.php" />
-            <file name="tests/Doctrine/Tests/Common/Proxy/ProxyGeneratorTest.php" />
             <directory name="vendor" />
```
stof: this ignore rule is also useless AFAICT

SenseException (approved):

PR #978, opened 2022-08-23
Remove test suite bootstrap file

greg0ire:
It does not appear to serve any purpose.


malarzm (approved):

SenseException (approved):
The last changes with functionality in TestInit was years ago and the tests don't break. It might be connected to the files that were removed over time.

PR #985, opened 2022-09-18
Fix consts in default values

### LINKED ISSUE #983: 3.4.0 broke class constants defaults in methods parameters when using PHP 8.1+
Hello,

Looks like https://github.com/doctrine/common/commit/a45d43712344ae0f4e1554ddcab39cb5a995e4c3 broke default values using other other references in method parameters:
A class such as
```php

// App\AnotherNamespace\DataType class
<?php
namespace App\AnotherNamespace;

class DataType {
const DAY = 'day';
}


// App\Entity\MyEntity class
<?php
namespace App\Entity;

use App\AnotherNamespace\DataType;
class MyEntity {
public function doSomething(string $dataType = DataType::DAY) {
// Do something
}
```
With 3.3.1:
```php
// App\Entity\MyEntity class
<?php
class MyEntity extends \App\Entity\MyEntity implements \Doctrine\ORM\Proxy\Proxy
public function doSomething(string $dataType = 'day'): void {
}
```

With 3.4.0:
```php
// App\Entity\MyEntity class
<?php
class MyEntity extends \App\Entity\MyEntity implements \Doctrine\ORM\Proxy\Proxy
public function doSomething(string $dataType = App\AnotherNamespace\DataType::DAY): void {
}
```

Which leads to a `Class "App\AnotherNamespace\DataType" does not exist` error, as namespace imports aren't used, or neither global namespace is used.

malarzm:
Fixes #983. cc @nicolas-grekas I've added you as co-author :)

### DIFF
--- a/src/Proxy/ProxyGenerator.php
+++ b/src/Proxy/ProxyGenerator.php
@@ -44,11 +44,14 @@ use function method_exists;
 use function mkdir;
 use function preg_match;
 use function preg_match_all;
+use function preg_replace;
+use function preg_split;
 use function random_bytes;
 use function rename;
 use function rtrim;
 use function sprintf;
 use function str_replace;
+use function strpos;
 use function strrev;
 use function strtolower;
 use function strtr;
@@ -58,6 +61,7 @@ use function var_export;
 
 use const DIRECTORY_SEPARATOR;
 use const PHP_VERSION_ID;
+use const PREG_SPLIT_DELIM_CAPTURE;
 
 /**
  * This factory is used to generate proxy classes.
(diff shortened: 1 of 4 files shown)


derrabus (approved):

src/Proxy/ProxyGenerator.php
```diff
@@ -1124,6 +1128,17 @@ private function getParameterDefaultValue(ReflectionParameter $parameter)
 
         $value = rtrim(substr(explode('$' . $parameter->getName() . ' = ', (string) $parameter, 2)[1], 0, -2));
 
+        if (strpos($value, '\\') !== false || strpos($value, '::') !== false) {
+            $value = preg_split("/('(?:[^'\\\\]*+(?:\\\\.)*+)*+')/", $value, -1, PREG_SPLIT_DELIM_CAPTURE);
+            foreach ($value as $i => $part) {
+                if ($i % 2 === 0) {
+                    $value[$i] = preg_replace('/(?<![a-zA-Z0-9_\x7f-\xff\\\\])[a-zA-Z0-9_\x7f-\xff]++(?:\\\\[a-zA-Z0-9_\x7f-\xff]++|::)++/', '\\\\\0', $part);
```
greg0ire, line 1135: I think lines 1131, 1134 and 1135 warrant explanations/code comments, I can't say I understand what they do. I've tried dumping `$value` on line 1133 and it does not seem to do any splitting in the case of the constant. Also, removing the dot in the regex on line 1132 produces no effect.

greg0ire: @nicolas-grekas maybe you can help? What should the `preg_split` part produce? Why are we acting only on `even` items of the array it produces?

malarzm:
As agreed internally, this PR demonstrably fixes a bug so I'm going to merge and release.

Issue #988, opened 2022-09-27
Wrong proxy class generation with default values as constants referncing `self`


Lustmored:
Recent fix for constants in default parameter values broke proxy generation when constant is referenced as `self`.

Example entity code:
```
public function hasAccess(int $level = self::LEVEL_WORKER): bool
    {
        return $this->level >= $level;
    }
```

Generated proxy class:
```
public function hasAccess(int $level = \self::LEVEL_WORKER): bool
    {

        $this->__initializer__ && $this->__initializer__->__invoke($this, 'hasAccess', [$level]);

        return parent::hasAccess($level);
    }

```

Which in effect causes Fatal Error in PHP 8.1:
```
[27-Sep-2022 09:11:12 UTC] PHP Fatal error:  '\self' is an invalid class name in /app/var/cache/beta/doctrine/orm/Proxies/__CG__AppEntitySomeEntity.php on line 242
```

malarzm:
I'll start regretting introducing support for new in initializers... Sorry for the bug, we'll try to fix it hopefully this week

nicolas-grekas:
Here is the code we use in Symfony to dump default values:
https://github.com/symfony/symfony/blob/6.2/src/Symfony/Component/VarExporter/ProxyHelper.php#L318

Feel free to borrow it (it's written for PHP >= 8.1)

PR #989, opened 2022-09-27
Use var_export for scalar default values

### LINKED ISSUE #987: ProxyGenerator turns double quoted default values into single quoted default values
Using PHP 8.1 and doctrine/common 3.4.1

the method in my entity

```php
public function formatAsString(string $rowDelim = "\n"): string
{
}
```

is turned in to
```php
public function formatAsString(string $rowDelim = '\n'): string
{
}
```

in the proxy.

----

Fyi:

This is the version that 3.3.1 generates
```php
public function formatAsString(string $rowDelim = '
'): string
{
}
```

malarzm:
This PR will be good enough to cut a 3.4.2 release. It will:

- fix #987 (also changes will need to stay until https://github.com/php/php-src/issues/9622 is resolved)
- stop breaking existing code like mentioned in #988 and give us more time to deal with current approach's shenanigans

It is worth mentioning that we will NOT support correctly double-quoted strings in complex default values until a fix in PHP is present (or a workaround is found). Given people never really had a chance to start using such values, I believe this is OK.

@ampaze @Lustmored may I ask you to test your apps with this PR instead of released 3.4.1?

### DIFF
--- a/src/Proxy/ProxyGenerator.php
+++ b/src/Proxy/ProxyGenerator.php
@@ -36,6 +36,7 @@ use function in_array;
 use function interface_exists;
 use function is_callable;
 use function is_dir;
+use function is_scalar;
 use function is_string;
 use function is_writable;
 use function lcfirst;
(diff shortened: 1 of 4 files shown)


ampaze:
@malarzm PR works for me.

greg0ire (approved):

Lustmored:
@malarzm it also works for me. Good job finding the fix so quickly, thank you!

PR #991, opened 2022-09-28
Fix nullable union type identifier lazy load

deguif:
Currently using union type for identifiers method doesn't permit to lazy load the entity.

With this fix, only a union type for a single type and null is supported. `null` can be at first position or at last.

For example:

This will enable lazy-loading (which previously was not): 

```php
public function getId(): int|null;
public function getId(): null|int;
public function getId(): \stdClass|null;
public function getId(): RelativeNamespace\Class|null;
public function getId(): \AbsoluteNamespace\Class|null;
```

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -17,4 +17,4 @@ jobs:
     name: "PHPUnit"
     uses: "doctrine/.github/.github/workflows/continuous-integration.yml@1.1.1"
     with:
-      php-versions: '["7.1", "7.2", "7.3", "7.4", "8.0", "8.1"]'
+      php-versions: '["7.1", "7.2", "7.3", "7.4", "8.0", "8.1", "8.2"]'
--- a/phpstan.neon.dist
+++ b/phpstan.neon.dist
@@ -1,5 +1,5 @@
 parameters:
-    phpVersion: 80100
+    phpVersion: 80200
     level: 3
     paths:
         - src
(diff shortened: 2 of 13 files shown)


malarzm:
Quick question as I haven't parsed the regex in my head yet: isn't this true for any unions and intersections?

deguif:
> isn't this true for any unions and intersections?

Current implementation only generate lazy loading identifier proxy for a single return type, which can be null. Is your question about supporting something like that?

```php
public function getId(): int|string|...|null
{
     return $this->id;
}
```

For intersection they are not supported by the regex, do you mean it could be supported?

malarzm:
> For intersection they are not supported by the regex, do you mean it could be supported?

Yes. Instead of extra support for `|null` we could introduce full support for any and all unions (and intersections while we're at it). I understand that `|null` will be the most common union used but if it would turn out not to be a big deal, then awesome :)

deguif:
@malarzm I added support for union and intersection type.

malarzm:
@deguif that's awesome! Would you mind adding a test for intersection as well?

deguif:
PHPStan is complaining about the union with intersection type (for PHP 8.2). This should be fixed with the next release, as the reflection classes were updated to support union type with intersection type.

deguif:
@malarzm tests are green now.
PHPStan 1.8.7 was released today, which allows the test for PHP 8.2 for union type with intersection type to success.

malarzm:
Amazing, thanks you very much for changes and clean history! 🍻  I'll try to review/merge/release this week

malarzm (approved):

malarzm:
Thanks @deguif!

PR #993, opened 2022-09-29
Don't allow generating proxies for readonly classes

nicolas-grekas:
Readonly classes cannot be proxied, let's reject them early on before we start generating broken proxies for them.


malarzm (approved):

malarzm:
Thanks @nicolas-grekas, this for sure will be handy in near future!

PR #997, opened 2022-10-14
fix: proxy with BackedEnum integer on identifier

Gwemox:
If entity identifier return an int BackedEnum, Proxy Generator generate method with cast to int. But it's impossible to cast a BackedEnum to int.

Call `getId()` on proxy throw an error : 
`Warning: Object of class App\\Enum\\ProviderId could not be converted to int`

Original entity:
```php
    public function getId(): ProviderId
    {
        return $this->id;
    }
```

Generated proxy method **before fix**:
```php
    /**
     * {@inheritDoc}
     */
    public function getId(): \App\Enum\ProviderId
    {
        if ($this->__isInitialized__ === false) {
            return (int)  parent::getId();
        }


        $this->__initializer__ && $this->__initializer__->__invoke($this, 'getId', []);

        return parent::getId();
    }
```

Generated proxy method **after fix**:
```php
    /**
     * {@inheritDoc}
     */
    public function getId(): \App\Enum\ProviderId
    {
        if ($this->__isInitialized__ === false) {
            return  parent::getId();
        }


        $this->__initializer__ && $this->__initializer__->__invoke($this, 'getId', []);

        return parent::getId();
    }
```

### DIFF
--- a/src/Proxy/ProxyGenerator.php
+++ b/src/Proxy/ProxyGenerator.php
@@ -27,6 +27,7 @@ use function call_user_func;
 use function chmod;
 use function class_exists;
 use function dirname;
+use function enum_exists;
 use function explode;
 use function file;
 use function file_put_contents;
@@ -940,7 +941,12 @@ EOT;
             if ($this->isShortIdentifierGetter($method, $class)) {
                 $identifier = lcfirst(substr($name, 3));
                 $fieldType  = $class->getTypeOfField($identifier);
-                $cast       = in_array($fieldType, ['integer', 'smallint'], true) ? '(int) ' : '';
+
+                if (ClassUtils::containsEnumType($method->getReturnType())) {
+                    $cast = '';
+                } else {
+                    $cast = in_array($fieldType, ['integer', 'smallint'], true) ? '(int) ' : '';
+                }
 
                 $methods .= '        if ($this->__isInitialized__ === false) {' . "\n";
                 $methods .= '            ';
--- a/src/Util/ClassUtils.php
+++ b/src/Util/ClassUtils.php
@@ -4,13 +4,20 @@ namespace Doctrine\Common\Util;
 
 use Doctrine\Persistence\Proxy;
 use ReflectionClass;
+use ReflectionIntersectionType;
+use ReflectionNamedType;
+use ReflectionUnionType;
 
+use function assert;
+use function enum_exists;
 use function get_class;
 use function get_parent_class;
 use function ltrim;
 use function rtrim;
 use function strrpos;
 use function substr;
+
+use const PHP_VERSION_ID;
 
 /**
  * Class and reflection related functionality for objects that
@@ -109,5 +116,32 @@ class ClassUtils
     public static function generateProxyClassName($className, $proxyNamespace)
     {
         return rtrim($proxyNamespace, '\\') . '\\' . Proxy::MARKER . '\\' . ltrim($className, '\\');
+    }
+
+    /**
+     * Check if the type is an enum type or type containing an enum type
+     *
+     * @param ReflectionNamedType|ReflectionUnionType|ReflectionIntersectionType|null $type
+     *
+     */
+    public static function containsEnumType($type): bool
+    {
+        if (PHP_VERSION_ID <= 80100 || $type === null) {
+            return false;
+        }
+
+        if ($type instanceof ReflectionUnionType || $type instanceof ReflectionIntersectionType) {
+            foreach ($type->getTypes() as $unionedType) {
+                if (self::containsEnumType($unionedType)) {
+                    return true;
+                }
+            }
+
+            return false;
+        }
+
+        assert($type instanceof ReflectionNamedType);
+
+        return enum_exists($type->getName());
     }
 }


src/Util/ClassUtils.php
```diff
@@ -110,4 +117,31 @@ public static function generateProxyClassName($className, $proxyNamespace)
...
         return rtrim($proxyNamespace, '\\') . '\\' . Proxy::MARKER . '\\' . ltrim($className, '\\');
     }
+
+    /**
+     * Check if the type is an enum type or type containing an enum type
+     *
+     * @param ReflectionNamedType|ReflectionUnionType|ReflectionIntersectionType|null $type
+     *
+     */
+    public static function containsEnumType($type): bool
```
stof, line 127: I would type the argument as `ReturnType|null`, to match the one of `ReflectionMethod::getReturnType`

src/Util/ClassUtils.php
```diff
@@ -110,4 +117,31 @@ public static function generateProxyClassName($className, $proxyNamespace)
...
+    {
+        if (PHP_VERSION_ID <= 80100 || $type === null) {
+            return false;
+        }
+
+        if ($type instanceof ReflectionUnionType || $type instanceof ReflectionIntersectionType) {
+            foreach ($type->getTypes() as $unionedType) {
+                if (static::containsEnumType($unionedType)) {
```
stof: I would use `self::` here. there is no reason to enable late static binding.

src/Util/ClassUtils.php
Gwemox: Union types are only allowed since PHP 8.0.

The PHP min version for this project is 7.1.

src/Util/ClassUtils.php
Gwemox: Yes, It's corrected

src/Proxy/ProxyGenerator.php
```diff
@@ -940,7 +941,12 @@ private function generateMethods(ClassMetadata $class)
             if ($this->isShortIdentifierGetter($method, $class)) {
                 $identifier = lcfirst(substr($name, 3));
                 $fieldType  = $class->getTypeOfField($identifier);
-                $cast       = in_array($fieldType, ['integer', 'smallint'], true) ? '(int) ' : '';
+
+                if (ClassUtils::containsEnumType($method->getReturnType())) {
```
malarzm, line 945: I'm afraid this is only half-baked solution as without method's return type we still do not know what's inside. Maybe it'd be more efficient to generate an `instanceof \BackedEnum` check for the `return` statement?

src/Proxy/ProxyGenerator.php
```diff
@@ -940,7 +941,12 @@ private function generateMethods(ClassMetadata $class)
...
+                    $cast = '';
+                } else {
+                    $cast = in_array($fieldType, ['integer', 'smallint'], true) ? '(int) ' : '';
```
malarzm, line 948: Or better yet: idea that needs somebody with a deeper ORM knowledge than me to be validated: what if we get rid of the cast instead? The proxied entity will have the field already casted to a correct type thanks to initial id hydration. This `$cast` construct here is 10 years old and removing it doesn't break any test so I think it's worth challenging.

malarzm (commented):
Whichever way we go, this PR will need test to prevent regressions in the future :)

src/Proxy/ProxyGenerator.php
Gwemox: Yes I think that's the best way.  

If the method is typed, for the scalar types PHP already cast return value if needed?

beberlei: @malarzm `AbstractProxyFactory::getProxy` is responsible for setting the id field on the proxy, its not gone through hydration, instead it might just be the value passed to `EntityManager::getReference($entity, $id)` without casting. So getting rid of the cast would require to move the cast to a higher level in the ORM. Same for UnitOfWork::createEntity using `getProxy`, associated ids don't have types and are not converted to int for example.

A test in ORM would be:

```php
$proxy = $em->getReference(CmsUser, "1");
$proxy->getId();
```
Unsure if we test this at the moment.

malarzm: Thanks @beberlei I'll try to have a closer look there in ORM 👍

Gwemox: @malarzm any news on this?

malarzm: Quoting myself from #998 

> Probably we won't end with this exact solution as it seems ORM has no tests for this behaviour either. The change looks OK ORM-wise, but I'm afraid it may hit users in an unexpected and hard to debug way.

Sadly I didn't have time to get back to this :(

Gwemox: How can I help ?

src/Proxy/ProxyGenerator.php
Gwemox: Check instance in template, like this ?
from:
```   
        if ($this->__isInitialized__ === false) {
            return (int)  parent::getId();
        }
```

to:
```   
        if ($this->__isInitialized__ === false) {
            return  parent::getId() instanceof \BackedEnum ? parent::getId() : (int) parent::getId();
        }
```

PR #1002, opened 2023-07-12
Deprecate all symbols in the Doctrine\Common\Proxy namespace

nicolas-grekas:
Allowed by https://github.com/doctrine/orm/pull/10837

### DIFF
--- a/src/Proxy/Autoloader.php
+++ b/src/Proxy/Autoloader.php
@@ -21,6 +21,7 @@ use const DIRECTORY_SEPARATOR;
  * Special Autoloader for Proxy classes, which are not PSR-0 compliant.
  *
  * @internal
+ * @deprecated The Autoloader class is deprecated since doctrine/common 3.5.
  */
 class Autoloader
 {
--- a/src/Proxy/ProxyGenerator.php
+++ b/src/Proxy/ProxyGenerator.php
@@ -67,6 +67,8 @@ use const PREG_SPLIT_DELIM_CAPTURE;
 /**
  * This factory is used to generate proxy classes.
  * It builds proxies from given parameters, a template and class metadata.
+ *
+ * @deprecated The ProxyGenerator class is deprecated since doctrine/common 3.5.
  */
 class ProxyGenerator
 {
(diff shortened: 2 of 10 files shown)


src/Proxy/ProxyGenerator.php
```diff
@@ -259,6 +262,14 @@ public function __getLazyProperties(): array
      */
     public function __construct($proxyDirectory, $proxyNamespace)
     {
+        if (PHP_VERSION_ID >= 80100) {
```
nicolas-grekas: version check matching https://github.com/doctrine/orm/pull/10837#discussion_r1261234947

derrabus:
Do we actually plan to release a v4 without Proxies? If the ORM does not use common proxies anymore, we'd simply drop the dependency.

nicolas-grekas:
> Do we actually plan to release a v4 without Proxies?

This would cut the maintenance yes. Proxies are hard to maintain, especially in the form they take in this repo (see PRs related to readonly or enums for examples.)

derrabus:
I understand that, but what remains of this library if we remove proxies is not worth maintaining imho. I think, we can simply archive the repo once the ORM doesn't use it anymore.

nicolas-grekas:
It'd still be great to release a version with deprecations, so that ppl that rely on eg `instanceof Common\Proxy\Proxy` can know they should move on.

derrabus:
Okay, how about:

* Annotate _all_ classes and interfaces with `@deprecated`, not just the Proxy stuff
* Remove all mentions of version 4.0
* No runtime deprecation warnings as there is no urge to resolve them.
* A note in the README about our plans of sunsetting this library

nicolas-grekas:
>Remove all mentions of version 4.0
> No runtime deprecation warnings as there is no urge to resolve them.

Done for those

For the rest, that's a separate effort to me.

src/Proxy/Autoloader.php
```diff
@@ -20,7 +20,7 @@
 /**
  * Special Autoloader for Proxy classes, which are not PSR-0 compliant.
  *
- * @internal
+ * @deprecated The Autoloader class is deprecated since doctrine/common 3.5.
```
derrabus: Deprecating the class does not make it not internal. 😬

derrabus (dismissed):

src/Proxy/Autoloader.php
nicolas-grekas: fix :see_no_evil:

derrabus (approved):

PR #1005, opened 2023-08-08
ProxyGenerator: do not generate long file names

JanTvrdik:
The maximum file name length when using [eCryptfs](https://www.ecryptfs.org/) is 143 bytes. This causes issues with Doctrine proxies where with current implementation this limit can be exceeded.

This PR changes how proxy file name is generated to avoid hitting this limit.

> eCryptfs can only store filenames of up to 143 characters when filename encryption is enabled. The remaining 112 characters are used for storing metadata such as the encrypted filename prefix, the signature of the filename encryption key, and an identifier for the cipher used, as well as random bytes to make /foo/file and /bar/file encrypt to different ciphertext and padding bytes to achieve proper cipher block size alignment.
> https://bugs.launchpad.net/ecryptfs/+bug/344878

### DIFF
--- a/src/Proxy/Autoloader.php
+++ b/src/Proxy/Autoloader.php
@@ -28,7 +28,7 @@ class Autoloader
      * Resolves proxy class name to a filename based on the following pattern.
      *
      * 1. Remove Proxy namespace from class name.
-     * 2. Remove namespace separators from remaining class name.
+     * 2. Replace namespace separators with directly separators in remaining class name.
      * 3. Return PHP filename from proxy-dir with the result from 2.
      *
      * @param string $proxyDir
@@ -47,10 +47,10 @@ class Autoloader
         }
 
         // remove proxy namespace from class name
-        $classNameRelativeToProxyNamespace = substr($className, strlen($proxyNamespace));
+        $classNameRelativeToProxyNamespace = substr($className, strlen($proxyNamespace) + 1);
 
-        // remove namespace separators from remaining class name
-        $fileName = str_replace('\\', '', $classNameRelativeToProxyNamespace);
+        // replace namespace separators in remaining class name
+        $fileName = str_replace('\\', DIRECTORY_SEPARATOR, $classNameRelativeToProxyNamespace);
 
         return $proxyDir . DIRECTORY_SEPARATOR . $fileName . '.php';
     }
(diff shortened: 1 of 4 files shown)


derrabus:
We're currently sunsetting the Common library including the proxy implementation. This is why I'm a bit reluctant to introduce this kind of change into the codebase at this point.

Furthermore, this is not a bugfix but rather a workaround for a very arbitrary limitation of the filesystem that you are using.

Can you elaborate what you use proxies for? Have you tried to work around the issue without changing doctrine/common?

JanTvrdik:
> Can you elaborate what you use proxies for? 

I'm using proxies as part of doctrine/orm.

> Have you tried to work around the issue without changing doctrine/common?

Yes, but there is no easy way to use custom logic for naming proxies, because `ProxyGenerator` instance is created in constructor of `Doctrine\ORM\Proxy\ProxyFactory` and `ProxyFactory` instance is created in constructor of `Doctrine\ORM\EntityManager`.

stof:
@JanTvrdik on PHP 8.1 and newer, the ORM has a new proxy implementation (currently opt-in) that is based on the trait provided by symfony/var-exporter.

However, I think the new code in Doctrine\ORM\Proxy\ProxyFactory in doctrine/orm 2.17.x uses the same naming convention for files and so would suffer from the same issue.

stof:
@derrabus IMO, we should still do a bugfix in doctrine to avoid generating file names of arbitrary lengh. *All* filesystems have such limitations (it is just that eCryptfs has a max file name length that is but shorter than ext4 so it is easier to reach it). See https://bugs.launchpad.net/ecryptfs/+bug/344878/comments/10 for a comment in the ecryptfs explaining that.

The current file name generating takes the FQCN of the entity and uses that as part of the file name as is. This could even reach the ext4 limit if the FQCN is long enough.

However, I'm not sure spreading generated proxies in subfolder is the right approach to fix this.

derrabus:
> However, I think the new code in Doctrine\ORM\Proxy\ProxyFactory in doctrine/orm 2.17.x uses the same naming convention for files and so would suffer from the same issue.

Then let's start with the ORM please. ORM 2.17 with recommended settings will bypass whatever we change in doctrine/common anyway.

PR #1012, opened 2024-04-16
explicitly mark nullable parameters as nullable

xabbuh:


derrabus:
I'm confused. Didn't #1011 fix that already?

xabbuh:
indeed, I missed that Nicolas fixed this after I forked the repo

PR #1022, opened 2024-12-30
Add doctrine/persistence v4 support

toniperic:


greg0ire:
Hi and thanks for working on this. May I ask why you're doing this though? The plan (although I don't think it's communicated clearly yet) is to retire and abandon `doctrine/common`. So… do you have a need that prompted this PR?

toniperic:
Hi @greg0ire, thanks for the notice. Had no idea that was the case.

As for the motivation: a project of mine depends on both `doctrine/common` and `doctrine/persistence` _(currently v3)_. 

Since `doctrine/persistence` [v4 got released roughly ~2 months ago](https://packagist.org/packages/doctrine/persistence#4.0.0), I've wanted to upgrade to it, but Composer has been flagging dependency resolution issues as `doctrine/common` works only with v2 and v3, which makes `doctrine/persistence` v4 impossible to install.

Considering how easy of an upgrade it was to get `doctrine/common` to work with v4, decided to give it a go.

greg0ire (approved):

greg0ire:
Re-targeted to 3.5.x as this is not a bugfix. I think we should release 3.5.0 shortly after merging, it contains https://github.com/doctrine/common/pull/1002, which will improve communication about deprecated stuff you might be using.

SenseException (approved):

